PR #4, opened 2022-10-18
Repository API design and implementation

markpatton:


pass-core-main/src/test/java/org/eclipse/pass/object/IntegrationTest.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright 2019, Yahoo
```
jabrah, line 2: Should this license statement be changed to our project default statement?

jabrah (approved):
Not too much from my perspective outside of the open question about the filter implementation and some minor comment editing. I'm amused by some of the seemingly gory Elide detail, like the to_json_api_doc method and tx handling

I do prefer the idea of the client impl triggering hooks and security checks as expected, but I'm not entirely sure if those come any usage caveats

pass-core-object-service/src/main/java/org/eclipse/pass/object/RSQL.java
```diff
@@ -0,0 +1,99 @@
...
+
+    public static String equals(String name, String value) {
+        return comparison(name, "==", value);
+    }
+
+    public static String notEquals(String name, String value) {
+        return comparison(name, "!=", value);
+    }
+
+    public static String in(String name, String... values) {
```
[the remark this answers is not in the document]
markpatton: Good question. My thought was to wait and then respond to user needs. There are also a bunch of questions about full text search to think about. Using glob patterns is probably a bad idea.

pass-core-main/src/test/java/org/eclipse/pass/object/IntegrationTest.java
markpatton: Right. Missed that.

PR #5, opened 2022-10-18
Update server config

jabrah:
* Update CORS to allow `origin: *`
* Add base URL as environment variable to force `https` in links. Ideally this behavior will change over to using relative URLs, but this should be acceptable for now
  * Add a default value in `.env` and a note in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,8 +45,8 @@ Environment variables:
 * PASS_CORE_DATABASE_PASSWORD=moo
 * PASS_CORE_PORT=8080
 * PASS_CORE_JAVA_OPTS="-Djavax.persistence.schema-generation.database.action=create"
-* PASS_CORE_BASE_URL=http://localhost
-  * Used when building relationship links
+* PASS_CORE_BASE_URL
+  * Used when building relationship links. This property does not have a default value and must be defined in your environment. The `pass-core-main/.env` is intended to be used for local testing of pass-core in isolation. If we want to use this in the local PASS demo environment, for example, we would specify `PASS_CORE_BASE_URL=https://pass.local`
 
 
 # Using JSON API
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -45,6 +45,8 @@ Environment variables:
 * PASS_CORE_DATABASE_PASSWORD=moo
 * PASS_CORE_PORT=8080
 * PASS_CORE_JAVA_OPTS="-Djavax.persistence.schema-generation.database.action=create"
+* PASS_CORE_BASE_URL=http://localhost
```
aforward, line 48: Should it be `https://localhost` to avoid the 302 redirect?

```
a2forward@demo:~$ curl -k http://localhost
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>302 Found</title>
</head><body>
<h1>Found</h1>
<p>The document has moved <a href="https://localhost/">here</a>.</p>
</body></html>
```

Vesus 

```
a2forward@demo:~$ curl -k https://localhost
<html>

<head>
    <title>PASS docker</title>
</head>

<body>
    PASS Docker!

    <p>Apps:</p>
    <ul>
        <li><a href="/app/">Ember - the PASS UI</a></li>
    </ul>

</body>

</html>
```

aforward (commented):
Consider `https` but otherwise looks great.

jabrah:
Should probably be clearer, the env var doesn't have a default value, instead the `.env` is used in the project's `docker-compose`. Presumably the repo will have integration tests run off this setup, but for now it's unused, other than to build the Docker image. In deployed environments, it would be expected that this var is present (I don't actually know how it would behave it it is missing)

markpatton (approved):
I ran the update in the jar and the docker environment and it looks good.

README.md
jabrah: Updated the README to try to better capture the intent of this value

PR #7, opened 2022-10-27
Unit Test for Model Objects  for classes in org.eclipse.pass.object.model

jrmartino:
We adapted the old test classes from pass-model in the pass-java-client project. Since we no longer need the JSON <-> object (de)serialization, these test methods were omitted. The new tests reflect changes made to the model, especially adapted for use with our new storage regime.


markpatton (commented):
We have an incoming pr that does some test setup with a newer version of junit. Let's get that in and then update this to the newer junit

jrmartino:
made changes to the unit tests to use jupiter, looks clean

pass-core-object-service/src/test/java/org/eclipse/pass/object/model/support/Identifier.java
```diff
@@ -0,0 +1,111 @@
+/*
```
markpatton, line 1: Should not be needed.

pass-core-object-service/src/test/java/org/eclipse/pass/object/model/support/IdentifierTests.java
```diff
@@ -0,0 +1,80 @@
+/*
```
markpatton, line 1: Not needed.

markpatton (commented):
I don't think the Identifier and Identifier test classes are needed. Otherwise looks good.

pass-core-object-service/src/test/java/org/eclipse/pass/object/model/support/Identifier.java
jrmartino: right you are, have pushed deletion of these files

markpatton (approved):
Looks good now.

PR #8, opened 2022-10-31
initial commit of pass doi service to pass-core

jrmartino:
This is an initial import of the doi service code into pass-core. The code has been refactored to separate communication with xref from communication with Elide.


pass-core-doi-service/README.md
```diff
@@ -0,0 +1,24 @@
+# pass-doi-service
+
+Service for accepting a DOI and returning a Journal ID and Crossref metadata for the DOI
+
+## Description
+
+This service accepts a journal DOI as a query parameter:
+
+`http://<host>:<port>/pass-doi-service/journal?doi=<doi>`
```
markpatton, line 9: This should be updated to be something like /journal.

pass-core-doi-service/pom.xml
```diff
@@ -0,0 +1,79 @@
...
+
+    <dependencies>
+        <dependency>
+            <groupId>org.eclipse.pass</groupId>
+            <artifactId>pass-core-object-service</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.yahoo.elide</groupId>
+            <artifactId>elide-spring-boot-starter</artifactId>
```
markpatton, line 29: This should not be a dependency.

pass-core-doi-service/pom.xml
```diff
@@ -0,0 +1,79 @@
...
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <version>${junit.jupiter.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>3.9.0</version>
```
markpatton, line 74: Should make this version a property.

pass-core-doi-service/src/main/java/org/eclipse/pass/doi/service/ElideConnector.java
```diff
@@ -0,0 +1,297 @@
...
+        Journal passJournal = find(name, issns);
+
+        //create or update the pass version of this Journal
+        if (passJournal == null) {
+            // we don't have this journal in pass yet
+            if (name != null && !name.isEmpty() && issns.size() > 0) {
+                // but we have enough info to make a Journal entry
+                try (ElideDataStorePassClient passClient = getNewClient()) {
+                    passClient.createObject(journal);
+                    passJournal = new Journal(find(name, issns));
```
markpatton, line 178: Is there a need for the copy Constructor here? 
Should try not to nest the PassClient usage.
Probably simplest to just create one PassClient and use throughout the method.
Another way of structuring things is just create one PassClient for each servlet request and use it throughout

pass-core-doi-service/src/main/java/org/eclipse/pass/doi/service/ElideConnector.java
```diff
@@ -0,0 +1,297 @@
...
+public class ElideConnector {
+    private static final Logger LOG = LoggerFactory.getLogger(ElideConnector.class);
+
+    protected RefreshableElide refreshableElide;
+
+    protected ElideConnector(RefreshableElide refreshableElide) {
+        this.refreshableElide = refreshableElide;
+    }
+
+    protected ElideDataStorePassClient getNewClient() {
```
markpatton, line 49: Should return PassClient not implementation class

pass-core-doi-service/src/main/java/org/eclipse/pass/doi/service/ElideConnector.java
```diff
@@ -0,0 +1,297 @@
...
+     * @return the URI of the best match, or null in nothing matches
+     */
+    protected Journal find(String name, List<String> issns) {
+
+        //keep track of hits for searches
+        List<Journal> foundList = new ArrayList<>();
+
+        //look for journals with this name
+        try (ElideDataStorePassClient passClient = getNewClient()) {
+            String filter = RSQL.equals("journalName", name);
```
markpatton, line 221: Is exact string matching what we want here? Maybe at least case insensitive?

pass-core-doi-service/src/main/java/org/eclipse/pass/doi/service/PassDoiServiceController.java
```diff
@@ -0,0 +1,150 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(PassDoiServiceController.class);
+    ElideConnector elideConnector;
+    XrefConnector xrefConnector;
+
+    PassDoiServiceController(RefreshableElide refreshableElide) {
+        this.elideConnector = new ElideConnector(refreshableElide);
+        this.xrefConnector = new XrefConnector();
+    }
+
+    @GetMapping("pass-doi-service/journal")
```
markpatton, line 44: Let's try for something just like /journal for the moment and see what things look like as all the other services get added.

pass-core-main/src/test/java/org/eclipse/pass/doi/service/DoiServiceTest.java
```diff
@@ -0,0 +1,216 @@
...
+import org.eclipse.pass.object.model.Journal;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+
+public class DoiServiceTest extends IntegrationTest {
+
+    @Autowired
+    protected RefreshableElide refreshableElide;
+
+    protected ElideDataStorePassClient getNewClient() {
```
markpatton, line 33: Should return PassClient and not implementation class

pass-core-main/src/test/java/org/eclipse/pass/doi/service/DoiServiceTest.java
```diff
@@ -0,0 +1,216 @@
...
+    public void realJournalTest() throws Exception {
+        String name = "Clinical Medicine Insights: Cardiology";
+        HttpUrl url = formDoiUrl("10.4137/cmc.s38446" );
+        String id;
+
+        //if this journal is in the database already, delete it
+        try (ElideDataStorePassClient passClient = getNewClient()) {
+            String filter = RSQL.equals("journalName", name);
+            PassClientResult<Journal> result = passClient.
+                selectObjects(new PassClientSelector<Journal>(Journal.class, 0, 100, filter, null));
```
markpatton, line 139: You could use streamObjects to make this simpler

markpatton (changes requested):
Looking good. Just a few cleanups suggested.

pass-core-doi-service/README.md
jrmartino: we are currently listening at /pass-doi-service/journal in core to proess the Request - i was thinking that as we add more services to core, it would be good to have service endpoints identify the services they belong to. there might in the future be another endpoint for another core service which also deals with journal.

the above URL is what is run in the integrtion test

pass-core-doi-service/pom.xml
jrmartino: agree - removing

pass-core-doi-service/pom.xml
jrmartino: actually, removing, not used

pass-core-doi-service/src/main/java/org/eclipse/pass/doi/service/ElideConnector.java
jrmartino: i was using the copy constructor because the objects pulled in by the client are not available once the client closes. i can refactor to use just one client in the method

pass-core-doi-service/src/main/java/org/eclipse/pass/doi/service/ElideConnector.java
jrmartino: PassClient is private to the object-service package, not available here. This reflects the choice of implementation we talked about.

pass-core-doi-service/src/main/java/org/eclipse/pass/doi/service/ElideConnector.java
jrmartino: These names are pretty stable - coming either from Medline, PMC or Crossref.  Case issues would arise from an error somewhere. The more robust search would come from the issns on the journals, but we don't have that search functionality yet.

We could move to case insensitive.

pass-core-doi-service/src/main/java/org/eclipse/pass/doi/service/PassDoiServiceController.java
jrmartino: I will make the change, but think this is better :-)

pass-core-main/src/test/java/org/eclipse/pass/doi/service/DoiServiceTest.java
jrmartino: Interface is not availabe outside of object-service - do we want to make it public?

pass-core-doi-service/src/main/java/org/eclipse/pass/doi/service/ElideConnector.java
jrmartino: have refactored this method to not use copy constructor, and just one client. have also refactored the IT to use jut one clent in the real journal test.

pass-core-doi-service/src/main/java/org/eclipse/pass/doi/service/ElideConnector.java
markpatton: PassClient should work. I just tested.

pass-core-main/src/test/java/org/eclipse/pass/doi/service/DoiServiceTest.java
markpatton: It is already public. I was able to use it.

jrmartino: ok, done

pass-core-doi-service/src/main/java/org/eclipse/pass/doi/service/ElideConnector.java
jrmartino: done

markpatton (changes requested):
There are a few places in ElideConnect which are printing stack traces, which should really be logging.

Other than that, there is a lot of opening and closing off PassClient which does not seem to be needed. Generally you should be able to get a PassClient at the start of the request and close it when the response is made.

markpatton (approved):
Looks good!

PR #10, opened 2022-11-08
Add GitHub Actions workflow to update docker image used in pass-docker when pass-core is updated

grant-mcs:


aforward (approved):

.github/workflows/update-image.yml
```diff
@@ -0,0 +1,51 @@
...
+          repository: eclipse-pass/pass-docker
+          token: ${{ secrets.pass_docker_pat }}
+      - uses: oleksiyrudenko/gha-git-credentials@v2-latest
+        with:
+          token: ${{ secrets.pass_docker_pat }}
+      - name: Get pom version
+        id: pom_version
+        run: echo "POM_VERSION=${{ github.event.inputs.pomVersion }}" >> $GITHUB_OUTPUT
+      - name: Update nightly server config
+        run: python tools/update-image.py eclipse-pass.nightly.yml pass-core "${{ steps.pom_version.outputs.POM_VERSION }}@${{ steps.image_hash.outputs.IMAGE_HASH }}"
```
jabrah, line 46: The image in `eclipse-pass.nightly.yml` should be called `pass-core-main`

There's a reference in [`eclipse-pass.nightly.yml` that references `pass-core`](https://github.com/eclipse-pass/pass-docker/blob/main/eclipse-pass.nightly.yml#L13) which is spiking new builds. Looks like this is coming from this Action

jabrah (changes requested):

.github/workflows/update-image.yml
grant-mcs: Good point! I've updated this call to include the full path to the image instead of letting it use the default which assumes the image name matches the service name.

jabrah (approved):
✅

PR #13, opened 2022-11-14
Log into GHCR before trying to upload Docker image to it

grant-mcs:


.github/workflows/update-image.yml
```diff
@@ -20,6 +20,8 @@ jobs:
         uses: actions/setup-java@v1
         with:
           java-version: 11
+      - name: "Log into Docker"
+        run: echo ${{ secrets.pass_docker_pat }} | docker login ghcr.io --username grant-mcs --password-stdin
```
grant-mcs, line 24: Not sure how to handle the necessity of having an account attached to this. The only generic account we have in eclipse-pass is eclipsewebmaster which doesn't seem like the right one for something like this.

aforward (approved):

PR #17, opened 2022-12-16
350 FileService

tsande16:
Performing a maven clean install, the JAR runs and will run successfully in docker. The POST /file, GET /file/{id} and DELETE /file/{id} endpoints are working both run locally and in docker. The `FileSystemStorageService` unit tests should all pass. The S3 implementation has been initialized but no unit tests are complete, nor will they run locally/docker. The `FileSystemStorageService` _should_ run locally on Windows and Linux. 

NOTE: There was an accidental replace all for "file" -> "fileStorage". I noticed this today and they all should be reverted now.
NOTE: Files should be committed with Unix end of line. If this isn't the case let me know and I will check my git config global again.


tsande16:
The CI unit & integration tests failed. These reason for this is because in my unit test I'm actually trying to write to the local file system. I suppose I should mock the file system? What are you thought for this @markpatton @jrmartino

pass-core-file-service/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+        <artifactId>pass-core</artifactId>
+        <groupId>org.eclipse.pass</groupId>
+        <version>0.2.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>pass-core-file-service</artifactId>
+    <version>0.2.0-SNAPSHOT</version>
+
+    <properties>
+        <maven.compiler.source>11</maven.compiler.source>
```
markpatton, line 17: Some of these properties may already be defined in parent pom or eclipse root pom or should be moved there.

pass-core-file-service/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+        <maven.compiler.target>11</maven.compiler.target>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <ocfl.java.core>1.5.0</ocfl.java.core>
+        <ocfl.java.aws>1.5.0</ocfl.java.aws>
+        <springframework.spring.web>5.3.12</springframework.spring.web>
+        <google.code.gson>2.9.1</google.code.gson>
+        <liquibase.core>4.8.0</liquibase.core>
+    </properties>
+
+    <dependencyManagement>
```
markpatton, line 27: I don't think this pom import is needed. Instead should just add a dependency to pass-core-object-service

pass-core-file-service/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+            <version>${javax.json.api.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-web</artifactId>
+            <version>${springframework.spring.web}</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
```
markpatton, line 72: Note both gson and javax.json dependencies. Could only one be used?

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/StorageService.java
```diff
@@ -0,0 +1,41 @@
...
+ *   limitations under the License.
+ */
+package org.eclipse.pass.file.service.storage;
+
+import java.util.Optional;
+
+import org.springframework.core.io.ByteArrayResource;
+import org.springframework.web.multipart.MultipartFile;
+
+public interface StorageService {
```
markpatton, line 24: Need some Javadoc. Should standardize method names and terminology. For example could use File as the thing being operated on. Method names might be storeFile, deleteFile, getFile, etc.

Also note no checked exceptions mentioned. For example FileServiceException is defined.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
```diff
@@ -0,0 +1,139 @@
...
+        LOG.info("Pass File Service: Uploading New File");
+        try {
+            if (file.getBytes().length == 0) {
+                return ResponseEntity.badRequest().build();
+            }
+        } catch (IOException e) {
+            LOG.error("File Service: Error processing file upload: " + e);
+            return ResponseEntity.notFound().build();
+        }
+        storageService = storageFactory.createStorage(storageConfiguration.getStorageProperties());
```
markpatton, line 84: This should probably not be done per request. The storage service should be created in the servlet initialization.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/StorageProperties.java
```diff
@@ -0,0 +1,59 @@
...
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+package org.eclipse.pass.file.service.storage;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties(prefix = "spring.file-service")
```
markpatton, line 21: Need javadoc

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/StorageService.java
```diff
@@ -0,0 +1,41 @@
...
+    void init();
```
markpatton, line 25: Is there a need for init? Can this just happen in implementation constructors?

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/FileSystemStorageService.java
```diff
@@ -0,0 +1,249 @@
...
+
+        try {
+            if (mFile.isEmpty() || mFile == null) {
+                throw new FileSystemException("File Service: Files is empty or missing.");
+            }
+            String origFileName = mFile.getOriginalFilename();
+            String fileId = generateId();
+            String mimeType = URLConnection.guessContentTypeFromName(origFileName);
+
+            Path tempFileDir = this.tempLoc.resolve(Paths.get(origFileName))
```
markpatton, line 128: What if multiple people upload a file with the same name? Look at Files.createTempFile

markpatton (commented):
For StorageService, I think the methods should throw a checked exception. You could use the new exception you have defined or perhaps the existing IOException. This would simplify the exception handling over all and reduce the need for all the runtime exceptions being thrown.

There are two implementations of StorageService, one for the file system and one for s3, but both use the same library. Could there just be one implementation that is configured differently?

Also you should generally make sure to do work per request when it really only needs to be done once on service initialization.

pass-core-file-service/pom.xml
tsande16: Will do.

pass-core-file-service/pom.xml
tsande16: Will do.

pass-core-file-service/pom.xml
tsande16: Yes, this needs to be cleaned up.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/StorageService.java
tsande16: For this draft, my javadoc is incomplete. Since some things were changing I was going to wait. I will go ahead and complete the javadoc for this PR.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/StorageProperties.java
tsande16: Will do.

pass-core-file-service/pom.xml
tsande16: Yes, I did a factory pattern since I wasn't sure of what type of implementations we will have in the future and this would enable additions that would only require a new class rather than potentially making modifications to existing classes. I will refactor and redesign this to be more simple.

I did notice a lot of overlap of functionality, especially using the OCF library really simplified implementation. 

I agree with removing the custom exception. I thought it would be more useful in my initial design, but after implementation it became frustrating and I agree it would be better to simplify and use IOexception etc.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/StorageService.java
tsande16: I had done it this way for reasons in the future for static initializers. I thought it would be more future proof. 

Given your comment about simplifying the StorageService design this won't be an issue and I can use constructors.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/FileSystemStorageService.java
tsande16: Okay, I will use this. I have it appending a timestamp with ms in the temp storage, but using an existing method is better.

jrmartino (commented):
Noy much to add to Mark's comments, looks pretty good to me. One small thing would be to amend the copyright [rofile to reference $today,year ibstead of the hard-coded 2019 :-)

pass-core-file-service/pom.xml
tsande16: @markpatton when I do this it doesn't find my package import: `"package liquibase.repackaged.org.apache.commons.lang3 does not exist"`

When I look at the POM for the object-service I don't see a dependency for liquibase, but I see it in the POM for pass-core-main. Would you like for me to add that as a dependency?

pass-core-file-service/pom.xml
tsande16: - Simplified the exception handling, most cases are using `IOException` or `FileSystemException`
- Refactored the design, no longer using StorageFactory and both S3 and FileSystem are handled in 1 class: `FileStorageService`
- Initialization is handled via `BeanInitialization`

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
tsande16: - Done

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/FileSystemStorageService.java
tsande16: -Done, using the system temp directory now.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/StorageService.java
tsande16: - Added javadoc 
- Standardized method names
- Added checked exceptions

markpatton:
IOException might make more sense to throw than IOException for the FileStorageService.

For easier configuration, there could be one directory provided which the service would use for all of its file storage needs. The OCFL root, working, and temp directories could be created under that directory. Could also automatically set that directory to a newly created subdirectory of the system temp directory when it is not explicitly set.

The FileStorageService tests look like integration tests and should be moved to pass-core-main. That would probably let you remove the spring import and some other dependencies from the file service pom.

The default configuration should be file system using a temp dir.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
```diff
@@ -0,0 +1,147 @@
...
+    @PostMapping("/file")
+    public ResponseEntity<?> fileUpload(@RequestParam("file") MultipartFile file) {
+        StorageFile returnStorageFile;
+        try {
+            if (file.getBytes().length == 0) {
+                return ResponseEntity.badRequest().build();
+            }
+        } catch (IOException e) {
+            LOG.error("File Service: Error processing file upload: " + e);
+            return ResponseEntity.notFound().build();
```
markpatton, line 79: 400 bad request would probably make more sense as an error for uploading an empty file, assuming we want to disallow that.
5xx error codes should be used when the server has had a problem

All of the error codes used should be documented.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        try {
+            returnStorageFile = fileStorageService.storeFile(file);
+        } catch (IOException e) {
+            LOG.error("File Service: Error storing file upload: " + e);
+            return ResponseEntity.notFound().build();
+        } catch (NullPointerException e) {
```
markpatton, line 87: It's a little odd to be directly catching a NPE. What's the intent?

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/StorageFile.java
```diff
@@ -0,0 +1,213 @@
...
+    private String mimeType;
+    private String storageType;
+    private Long size;
+    private String extension;
+
+    /**
+     * Default Constructor which initializes the StorageFile with default value for member variables
+     *
+     */
+    public StorageFile() {
```
markpatton, line 39: Do these values makes sense as defaults? If not can everything just be null or 0? If not do you need a default constructor?

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/StorageFile.java
```diff
@@ -0,0 +1,213 @@
...
+                Objects.equals(this.extension, storageFile.extension);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(id, fileName, mimeType, storageType, size, extension);
+    }
+
+    @Override
+    public String toString() {
```
markpatton, line 190: This is a lot of work on toString! Maybe just use whatever the ide generates.

pass-core-main/src/main/resources/application.yaml
```diff
@@ -93,6 +101,14 @@ spring:
     driver-class-name: 'org.postgresql.Driver'
   liquibase:
     enabled: false
+  file-service:
+    storage-type: ${PASS_CORE_FILE_SERVICE_TYPE}
```
markpatton, line 105: Either the property code should handle the case of these not being set, or reasonable defaults should be provided here for when they are not set. You can do  `${ENV_VAR_NAME:default_value}`

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/FileStorageService.java
```diff
@@ -0,0 +1,362 @@
...
+        } else {
+            this.rootLoc = Paths.get(this.storageProperties.getStorageRootDir());
+        }
+
+        // The ocflLoc only needs to be set if the storage type is file system.
+        // If the storageType is S3 then workLoc and tempLoc are used because they are used with FILE_SYSTEM AND S3
+        if (storageType.equals(StorageServiceUtils.StorageServiceType.FILE_SYSTEM.label)) {
+            ocflLoc = Paths.get(this.rootLoc.toString(),this.storageProperties.getStorageOcflDir());
+        }
+        this.workLoc = Paths.get(this.rootLoc.toString(),this.storageProperties.getStorageWorkDir());
```
markpatton, line 124: It doesn't seem like the root storage subdirectory names don't need to be configurable.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/StorageServiceUtils.java
```diff
@@ -0,0 +1,79 @@
...
+ * @author Tim Sanders
+ */
+public final class StorageServiceUtils {
+
+    /**
+     * The StorageServiceType enum defines the type of storage supported by the File Service. The two types of
+     * persistence are supported: File Systems and S3 buckets. These values are to be used in the environment var
+     * configuration. If a new type of persistence is to be added, it must be added to this enum.
+     */
+    public enum StorageServiceType {
```
markpatton, line 36: This would work better as a toplevel enum.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/StorageProperties.java
```diff
@@ -0,0 +1,149 @@
...
+     */
+    public String getStorageTempDir() {
+        return tempDir;
+    }
+
+    /**
+     * Get the storage type that is configured for the File Service
+     * @return The storage type of the File Service
+     */
+    public String getStorageType() {
```
markpatton, line 84: If not going to use the enum here, no sense in really having it. Just define some static strings in this class to represent the different storage types.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/StorageServiceUtils.java
```diff
@@ -0,0 +1,79 @@
...
+        }
+        return sb.toString();
+    }
+
+    /**
+     * Helper method for the StorageFile class. Returns the absolute path of the file name.
+     * @param path The path to the file
+     * @return The absolute path of the file name
+     */
+    public static Path getAbsoluteFileNamePath(Path path) {
```
markpatton, line 75: Is this needed? What about `path.getParent()`?

markpatton (commented):
Added some comments on the code.

I notice if I try to run the jar without any configuration file upload fails and I get this message in the logs
```
 org.eclipse.jetty.server.HttpChannel.handleException - handleException /file java.io.IOException: File Service: File Service Type is missing or incorrect in the environment variables.
```
I'm gong to get it configured and look a little more at the behavior.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/StorageFile.java
tsande16: This is what was generated by IntelliJ. Would you like for me to condense this down?

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
tsande16: The NotFound is a 404. Would you like for me to make this BadRequest? 

I don't explicitly use 500 codes. I believe what you're seeing is from Springboot.

I will go ahead and document the error codes. I looked around to see how we documented but I don't see other examples. Do we have an example of how we want to document our HTTP error codes?

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
tsande16: I went back and forth on this, but I decided to do it this way so that it would return 404 instead of Springboot handling it and returning a 500.

pass-core-main/src/main/resources/application.yaml
tsande16: Will do!

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/FileStorageService.java
tsande16: I'm not sure if I understand this comment correctly, but for further clarification on my end, there is only one directory that is configurable and that is the root, the subdirectories which are no longer configurable are final member variables in the StorageProperties class.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/StorageFile.java
tsande16: I will make everything 0. Having an explicit default constructor at least from a springboot perspective is considered good form - This is just something I read years ago and stuck with it. The reasoning is that if an explicit default constructor is not defined, springboot can be unpredictable with it's bean initialization. I'm not an expert on springboot and perhaps this just might be a general rule from early days of springboot and things have changed. If you would for me to remove the default constructor I will do that.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/StorageProperties.java
tsande16: Yes, this is a good point, I will refactor this to use the enum.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/StorageServiceUtils.java
tsande16: I will remove this and use path.getParent()

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
tsande16: From our meeting: Any error responses should contain the exception message.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
tsande16: From our meeting: Add readme in the FileService that documents this as well.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/FileStorageService.java
tsande16: Resolved during meeting.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/StorageFile.java
tsande16: Resolved during meeting: not an issue.

jabrah:
Are you able to do a rebase against `main` to resolve the POM conflicts?

markpatton (changes requested):
Looking good. I am able to build and run the jar with the defaults and things works.

A few requests:

- Add the location where the file gets stored to the log messages.
- The Content-Type header should be set on retrieval. It's probably good enough just to use `Files.probeContentType`.
- The  Content-Disposition header should be set with the original filename on retrieval
- When the service starts out without a configured root directory and uses the system temporary directory, it should really use a random subdirectory under that. You can do this with ` Files.createTempDirectory`. This behavior should also be documented in the README.

tsande16:
Adding one more request from @markpatton - This was taken from our pass-dev slack channel:

Random idea about the file service, one way to deal with the original file name would be to have it be the last portion of the file identifier. As in the file identifier looks like RANDOM_STRING/ORIGINAL_FILE_NAME. Then file id would be a path rather than a path segment, but easily handled.

markpatton:
Another thought about the randomly generated identifiers. Currently it is generating a 25 character alpha numeric string. That is a huge number of bits! Maybe instead just generate a random uuid? See 'UUID.randomUUID()'. So as the file identifier you could use something like `UUID/ORIGNIAL_FILE_NAME` or if `/` isn't allowed, maybe separate with a `:`.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
```diff
@@ -0,0 +1,146 @@
...
+     * @return Bitstream The file requested by the fileId
+     */
+    @GetMapping("/file/{fileId:.+}")
+    @ResponseBody
+    public ResponseEntity<?> getFileById(@PathVariable String fileId) {
+        if (fileId == null) {
+            LOG.error("File ID not provided to get a file.");
+            return ResponseEntity.badRequest().body("File ID not provided to get a file.");
+        }
+        ByteArrayResource fileResource;
```
markpatton, line 106: This is a bit awkward. Assuming the file name can't have a - in it, you could just get the last -. Avoid regex when possible. This can also be done after the file is actually retrieved from the repository.

Also, - is actually pretty commonly used in filenames, it would be good to use another character if possible.

markpatton (commented):
I'm able to run it locally. I can upload files and retrieve. Headers are set appropriately. I made a couple more comments about the id handling in the code.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
tsande16: Okay, I will switch it to another character. I think you mentioned : before. Would you like for me to use this one instead?

tsande16: @markpatton decision from our slack: uuid / original filename in the path segement

PR #18, opened 2022-12-16
Add Unpaywall DOI Lookup to pass-doi-service module

jrmartino:
Some refactoring done here to more easily accomodate multiple doi-based services in the RestController.
Close attention needs to be paid to the form of returned JSON objects to the UI to be sure structure and escaping
are as expected by pass-ui.

### DIFF
--- a/pass-core-doi-service/README.md
+++ b/pass-core-doi-service/README.md
@@ -1,7 +1,7 @@
 # pass-doi-service
 
-Service for accepting a DOI and returning a Journal ID and Crossref metadata for the DOI via the /journal endpoint, or 
-returning information about a manuscript from Unpaywall via the /manuscript endpoint.
+Service for accepting a DOI and returning a Journal ID and Crossref metadata for the DOI via the /doi/journal endpoint, or 
+returning information about a manuscript from Unpaywall via the /doi/manuscript endpoint.
 
 ## Description for the `/doi/journal` endpoint
 
@@ -39,6 +39,6 @@ corresponding locations on the web for manuscript PDFs related to the article re
 
 ### Configuration
 
-Just like the /journal endpoint, this service will look for an environment variable called PASS_DOI_SERVICE_MAILTO 
-to specify a value for the `email` query parameter on the Unpaywall request. In addition, we may supply values for XREF_BASEURL
+Just as for the /doi/journal endpoint, this service will look for an environment variable called PASS_DOI_SERVICE_MAILTO 
+to specify a value for the `email` query parameter on the Unpaywall request. In addition, we may supply values for XREF_BASEURI
 and UNPAYWALL_BASEURI, which default to `https://api.crossref.org/v1/works/` and `https://api.unpaywall.org/v2/` respectively.
--- a/pass-core-doi-service/README.md
+++ b/pass-core-doi-service/README.md
@@ -26,7 +26,7 @@ header on the Crossref request. Default value os `pass@jhu/edu`.
 
 ## Description for the `/doi/manuscript` endpoint
 
-This service accepts a journal DOI as a query parameter:
+This service accepts a manuscript DOI as a query parameter:
 
 `http://<host>:<port>/doi/manuscript?doi=<doi>`
 
--- a/pass-core-doi-service/README.md
+++ b/pass-core-doi-service/README.md
@@ -3,11 +3,11 @@
 Service for accepting a DOI and returning a Journal ID and Crossref metadata for the DOI via the /journal endpoint, or 
 returning information about a manuscript from Unpaywall via the /manuscript endpoint.
 
-## Description for the `/journal` endpoint
+## Description for the `/doi/journal` endpoint
 
 This service accepts a journal DOI as a query parameter:
 
-`http://<host>:<port>/journal?doi=<doi>`
+`http://<host>:<port>/doi/journal?doi=<doi>`
 
 DOIs must contain a form like `10.1234/ ...`
 If a DOI is of a longer URL form containing the string `doi.org/`, then we truncate the DOI to take everything after
@@ -22,19 +22,19 @@ as a result of the Crossref call.
 ### Configuration
 
 The service will look for an environment variable called PASS_DOI_SERVICE_MAILTO to specify a value on the User-Agent
-header on the Crossref request. Default value os `pss@jhu/edu`.
+header on the Crossref request. Default value os `pass@jhu/edu`.
 
-## Description for the `/manuscript` endpoint
+## Description for the `/doi/manuscript` endpoint
 
 This service accepts a journal DOI as a query parameter:
 
-`http://<host>:<port>/manuscript?doi=<doi>`
+`http://<host>:<port>/doi/manuscript?doi=<doi>`
 
 DOIs must contain a form like `10.1234/ ...`
 If a DOI is of a longer URL form containing the string `doi.org/`, then we truncate the DOI to take everything after
 this substring.
 
-The service validates the form of the doi - if it is valid, then we hit the Onpaywall API to get information about the
+The service validates the form of the doi - if it is valid, then we hit the Unpaywall API to get information about the
 corresponding locations on the web for manuscript PDFs related to the article referenced by the DOI.
 
 ### Configuration
--- a/pass-core-doi-service/README.md
+++ b/pass-core-doi-service/README.md
@@ -1,8 +1,9 @@
 # pass-doi-service
 
-Service for accepting a DOI and returning a Journal ID and Crossref metadata for the DOI
+Service for accepting a DOI and returning a Journal ID and Crossref metadata for the DOI via the /journal endpoint, or 
+returning information about a manuscript from Unpaywall via the /manuscript endpoint.
 
-## Description
+## Description for the `/journal` endpoint
 
 This service accepts a journal DOI as a query parameter:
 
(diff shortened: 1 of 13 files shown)


markpatton:
I wonder about collecting the endpoints for this service together under /doi and giving them meaningful names. What about /doi/journal and /doi/manuscript? Do those names give a reasonable indication about what they are doing?

jaredgalanis:
> I wonder about collecting the endpoints for this service together under /doi and giving them meaningful names. What about /doi/journal and /doi/manuscript? Do those names give a reasonable indication about what they are doing?

This sounds good to me, but the one thing I just realized is that the manuscripts service is returning more than one manuscript so we could consider pluralizing the endpoint name. Keeping it singular for the sake of consistency also seems fine, just thought I'd point it out.

pass-core-doi-service/src/test/java/org/eclipse/pass/doi/service/UnpaywallDoiServiceTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public class UnpaywallDoiServiceTest {
+
+    ExternalDoiService underTest = new UnpaywallDoiService();
+
+    String unpaywallData =
```
jaredgalanis, line 29: This is the way the payload coming from the old download service looks:

```
{
  "manuscripts": [
    {
      "url": "https://europepmc.org/articles/pmc6759371?pdf=render",
      "repositoryLabel": "PubMed Central - Europe PMC",
      "type": "application/pdf",
      "source": "Unpaywall",
      "name": "pmc6759371?pdf=render"
    }
  ]
}
```

The payload described here looks like this in JSON format:

```
{
  "Manuscripts": [
      {
          "Location": "https://journals.sagepub.com/doi/pdf/10.4137/CMC.S38446",
          "RepositoryInstitution": null,
          "Type": "application/pdf",
          "Source": "Unpaywall",
          "Name": "CMC.S38446"
      },
      {
          "Location": "https://europepmc.org/articles/pmc5072460?pdf=render",
          "RepositoryInstitution": "PubMed Central - Europe PMC",
          "Type": "application/pdf",
          "Source": "Unpaywall",
          "Name": "pmc5072460?pdf=render"
      },
      {
          "Location": null,
          "RepositoryInstitution": null,
          "Type": "application/pdf",
          "Source": "Unpaywall",
          "Name": null
      }
  ]
}
```

This all seems reasonable to me. I think it's pretty much just the capitalization of the keys and the repository key phrasing that's changed. Both of those things would be easy to adapt to in the UI.

jrmartino: i'll make the changes to agree with current UI usage
thanks for the info!!

jrmartino: also - i wasn't a part of the discussion on the change from returning just the best oa_location to returningall of them, so i don't know why that switch was made. i am a little concerned that offering a choice without knowing which one is the best might lead users to select a pre-print version versus a published version for example, since the version info is not provided to assist the decision process

markpatton:
There are some Javadoc warnings that would be good to fix.

pass-core-doi-service/src/test/java/org/eclipse/pass/doi/service/UnpaywallDoiServiceTest.java
jaredgalanis: Thanks, @jrmartino! 

I think it was most likely related to the concern that even the best option might not be the one that a user was searching for / if there was issues downloading it.

Identifying the best options sounds like a decent improvement. If you include another property like `bestOaLocation: true` or something to that effect, we can easily make some changes to the UI to identify it. And it shouldn't break anything if you do and we haven't yet implemented messaging around it in the UI. 

That said, I don't want to hold this up and I could see it as a follow up issue too.

pass-core-doi-service/src/main/java/org/eclipse/pass/doi/service/ExternalDoiService.java
```diff
@@ -0,0 +1,149 @@
...
+/**
+ * ExternalDoiService classes provide configuration needed for specific implementations'
+ * connections, as well as a method to process the raw JSON object returned by the external service to
+ * suit the requirements of the PASS UI.
+ *
+ * @author jrm
+ */
+public abstract class ExternalDoiService {
+    private static final Logger LOG = LoggerFactory.getLogger(ExternalDoiService.class);
+    private final Set<String> activeJobs = new HashSet<>();
```
markpatton, line 40: Need to handle concurrent access to this set. Could do manual locking, use Collecitons.synchronizedSet or ConcurrentHashMap.

markpatton (commented):
Added a few comments on various files.

pass-core-doi-service/src/main/java/org/eclipse/pass/doi/service/ExternalDoiService.java
jrmartino: have addressed this, thanks

pass-core-doi-service/src/main/java/org/eclipse/pass/doi/service/ExternalDoiServiceConnector.java
```diff
@@ -0,0 +1,109 @@
...
+
+            return metadataJsonObject;
+        } catch (JsonParsingException e) {
+            if (responseString != null) {
+                return Json.createObjectBuilder()
+                           .add("error", responseString)
+                           .build();
+            }
+        } catch (IOException e) {
+            return null;
```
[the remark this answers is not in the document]
jrmartino: done, thanks

pass-core-doi-service/src/main/java/org/eclipse/pass/doi/service/UnpaywallDoiService.java
```diff
@@ -0,0 +1,87 @@
...
+public class UnpaywallDoiService extends ExternalDoiService {
+
+    @Override
+    public String name() {
+        return "Unpaywall";
+    }
+
+    @Override
+    public String baseUrl() {
+        String UNPAYWALL_BASEURI = "https://api.unpaywall.org/v2/";
```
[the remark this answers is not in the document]
jrmartino: done, thanks

pass-core-doi-service/src/test/java/org/eclipse/pass/doi/service/UnpaywallDoiServiceTest.java
jrmartino: easy enough to add, done - thanks for the suggestion. JSON element  is named "isBest" with a boolean value. this agrees with usage in the Unpaywall JSON. "bestOaLocation is too close to the JSON element best_oa_location" which refers to an entire Location (our "manuscript") entry.

pass-core-doi-service/README.md
```diff
@@ -18,7 +19,26 @@ corresponding journal. We then check to see if there is a
 containing the `journal-id` of the PASS journal, and a `crossref` object representing the data returned to the service
 as a result of the Crossref call.
 
-## Configuration
+### Configuration
 
 The service will look for an environment variable called PASS_DOI_SERVICE_MAILTO to specify a value on the User-Agent
-header on the Crossref request. 
+header on the Crossref request. Default value os `pss@jhu/edu`.
```
jabrah, line 25: Typo on email

pass-core-doi-service/README.md
```diff
@@ -18,7 +19,26 @@ corresponding journal. We then check to see if there is a
...
+
+This service accepts a journal DOI as a query parameter:
+
+`http://<host>:<port>/manuscript?doi=<doi>`
+
+DOIs must contain a form like `10.1234/ ...`
+If a DOI is of a longer URL form containing the string `doi.org/`, then we truncate the DOI to take everything after
+this substring.
+
+The service validates the form of the doi - if it is valid, then we hit the Onpaywall API to get information about the
```
jabrah, line 37: Small typo in this line `Onpaywall`

pass-core-doi-service/README.md
```diff
@@ -1,8 +1,9 @@
 # pass-doi-service
 
-Service for accepting a DOI and returning a Journal ID and Crossref metadata for the DOI
+Service for accepting a DOI and returning a Journal ID and Crossref metadata for the DOI via the /journal endpoint, or 
+returning information about a manuscript from Unpaywall via the /manuscript endpoint.
 
-## Description
+## Description for the `/journal` endpoint
```
jabrah, line 6: Does endpoint references need to be updated? 
`/doi/journal`, `/doi/manuscript`?

jabrah (commented):
Do the unit tests in `ExternalDoiServiceConnectorTest.java` make external web requests?
I'm a bit hesitant to have external requests in unit tests, but maybe I am misunderstanding the tests or maybe it's ultimately the best way

pass-core-doi-service/README.md
jrmartino: yes, good catch

pass-core-doi-service/README.md
jabrah: Updated to `pass@jhu/edu`, should be `pass@jhu.edu`

pass-core-doi-service/README.md
jabrah: For consistency, we should also update the URL path references on [lines 3-4](https://github.com/eclipse-pass/pass-core/blob/ce58419089154b93a67815bfa242ef0c8313b2c7/pass-core-doi-service/README.md?plain=1#L3-L4)

pass-core-doi-service/README.md
jrmartino: thanks, fixed

pass-core-doi-service/README.md
jrmartino: done

jabrah (approved):

tsande16 (approved):
Ran locally in docker and tested the endpoints in postman. Everything looks good and working as documented.

PR #21, opened 2023-01-11
organize submodule properties into module pom

jrmartino:


pass-core-doi-service/pom.xml
```diff
@@ -14,8 +14,6 @@
     <version>0.2.0-SNAPSHOT</version>
 
     <properties>
-        <maven.compiler.source>11</maven.compiler.source>
-        <maven.compiler.target>11</maven.compiler.target>
```
jabrah, line 18: Without these, will the project revert to the config in the parent POM in main? That one is configured to `1.8` for compatibility with the Fedora setup

markpatton: I think its ok for this release as long as it actually builds with it. First thing after the release we need to switch the main pom to 11.

jabrah (approved):

markpatton (approved):

pass-core-doi-service/pom.xml
jrmartino: build id ok. i am also adding in the spring plugin version property to get this to pass checks - we can remove after next release of pass-parent

PR #22, opened 2023-02-01
POM reorganization

jabrah:
* Remove parts that were added to most recent `eclipse-pass/main` POM
* Move some testing deps to top level POM


jrmartino (approved):
looks good, really nice cleanup, installs locally with no problems

PR #23, opened 2023-02-01
Add draft snapshot and release gh actions

jabrah:
Relates to https://github.com/eclipse-pass/main/issues/235

* Add new GitHub actions intended to publish to Sonatype. One for SNAPSHOT, separate one for full release
* SNAPSHOT publish will run on commit/merge in `main`
* Full release will run manually only, require input of release version and next dev version #'s

## Questions
@jrmartino Can you check the maven commands? 
* [SNAPSHOT command](https://github.com/eclipse-pass/pass-core/blob/235-sonatype-automation/.github/workflows/publish-snapshot.yml#L32)
* [Release commands](https://github.com/eclipse-pass/pass-core/blob/235-sonatype-automation/.github/workflows/release.yml#L42-L43)

I know the release commands are incomplete (minimum would require user input) and aren't yet setup to update sub-module versions correctly

## TODO
* Release action also incomplete, as we'd probably want the GH Action to push the new release tag to the GH repo, possibly create a new Docker image
* ~Both GH Actions require credentials in the form of GH secrets (https://docs.github.com/en/actions/security-guides/encrypted-secrets) which I do not have access to. Will need to poke someone else to handle this or open a ticket with Eclipse HelpDesk~ OSSHR_USERNAME and OSSRH_PASSWORD secrets are in place for `pass-core`. Other secrets will need to be added later

## Testing
Please focus on the `publish-snapshot` action. This will provide a PoC for handling Sonatype credentials. If the snapshots can get published with this action, then I can move onto making sure the full release can function.

There's no real way of testing this out until it's merged into `main`, at which point, we'll have a button to run the workflow(s) manually in the GH UI.

### DIFF
--- a/.github/workflows/publish-snapshot.yml
+++ b/.github/workflows/publish-snapshot.yml
@@ -6,11 +6,6 @@ on:
       - 'main'
   workflow_dispatch:
 
-# Will need to have someone with admin permissions to add some secrets:
-# Credentials for Sonatype, needs a Sonatype account. Use Jim's for testing?
-#   - OSSRH_USERNAME
-#   - OSSRH_TOKEN
-
 jobs:
   publish:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


jabrah:
The `release` action won't work because of missing GH secrets, but also shouldn't be run anyway. Must be manually triggered in the GH UI, so this won't be an issue.

The SNAPSHOT action should work, although I'm not sure if all of the secrets are setup correctly. For example, I'm not sure if I need GPG keys for artifact signing for SNAPSHOTs or if that is only required for full releases that will be synced to Maven Central

jrmartino (approved):
have not been able to test, but the maven commands ought to work as expected

PR #24, opened 2023-02-03
Added .gitattributes for LF line breaks

tsande16:
Docker requires entrypoint.sh and init_postgres.sh to have LF line breaks. Git on windows by default will checkout CRLF style and will cause Docker to bork.


jabrah (approved):
Should override the Windows default behavior without effecting git behavior in linux or mac systems. Worked for me in WSL just fine, though i wasn't able to clone a fresh copy of your repo

PR #25, opened 2023-02-06
408 - Added chmod to the entrypoint in Dockerfile

tsande16:
On a windows machine the execute permission gets removed during the maven image build. This error occurs after building an image from the maven process. 

Error response from daemon: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "./entrypoint.sh": permission denied: un
known

To remedy this issue a chmod +x is added to the Dockerfile. 

Tested locally in Docker. Pass-core run in docker and able to POST/GET a grant.

### DIFF
--- a/pass-core-main/Dockerfile
+++ b/pass-core-main/Dockerfile
@@ -1,11 +1,13 @@
 FROM openjdk:11-jre-slim
 
-RUN apt update \
-    && apt install -y curl
-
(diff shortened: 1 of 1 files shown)


jabrah (approved):

markpatton (approved):
Works for me locally on fedora linux

PR #26, opened 2023-02-07
Set pass-core-main image used by docker-compose.yml to 0.3.0-SNAPSHOT

markpatton:
Should also update release notes to set this value.


jabrah (approved):
👍

PR #27, opened 2023-02-09
Update snapshot publish action

jabrah:
## Changes

* Only run action for `*-SNAPSHOT` versions, as defined in the POM
* Update update-image action to accept secret for GHCR login user
* publish-snapshot will run on merge to `main`
* Should now work as intended
* GPG artifact signing should be configured and working now

Future work: 
* Promote secrets to organization level, so they can be used in `main` and `pass-support` (and wherever else necessary)


jrmartino (approved):
this looks like it should work

PR #29, opened 2023-02-16
Update release workflow

jabrah:
Related: https://github.com/eclipse-pass/main/issues/452

Needs testing. All secrets should be in place in `pass-core`


jrmartino (approved):
this looks reasonable to me

PR #30, opened 2023-02-21
User service and access control

markpatton:
Changes:
* Spring security enabled
  * Clients must authenticate with HTTP basic or by having appropriate Shib headers
  * HTTP basic for backend user in application.yaml
  * Shib users are automatically associated with a PASS user
    * PASS user is created and updated if needed
* PASS objects have authorization rules
* More sensible defaults in application yaml
  * Logging directory env variable
* Fix ElidePassClientTest not using correct class
* User.locatorIds and Journal.issns now searchable with has member RSQL operator
* Added user service
* Updated README

How to test:

First set the auth caching to 1 minute to make testing easier.
Edit `pass-core-main/src/main/resources/application.yaml` and set `pass.auth.cache-duration` to `1`.

Build and start pass-core:
```
mvn package
java -jar pass-core-main/target/pass-core-main.jar
```

Now all requests must either provide HTTP basic credentials as the backend user or set Shib headers as if they have been authenticated by Shib.

Make a request as the backend user:
```
curl -u backend:moo http://localhost:8080/data/grant
```

Check various services and verify that the backend user can make requests. Check that when credentials are not provided, requests return a 401.

Make a request as the Shib user:
```
curl http://localhost:8080/data/user/ -H "Eppn: sallysubmitter@johnshopkins.edu" -H "Displayname: Sally M. Submitter" -H "Mail: sally232@jhu.edu" -H "Givenname: Sally" -H "Sn: Submitter" -H "Affiliation: FACULTY@johnshopkins.edu" -H "Employeenumber: 02342342" -H "unique-id: sms2323@johnshopkins.edu"
```

Note that the request will return the newly created user. Check that the object looks request. Make the same request again and verify that the object remains the same. Wait for a minute and make a request with a different email or display name in the header. Verify that the existing object changes.

Try out the user service at `/user/whoami`. Just like all the other services, it needs the same header set.
```
curl http://localhost:8080/user/whoami -H "Eppn: sallysubmitter@johnshopkins.edu" -H "Displayname: Sally M. Submitter" -H "Mail: sally232@jhu.edu" -H "Givenname: Sally" -H "Sn: Submitter" -H "Affiliation: FACULTY@johnshopkins.edu" -H "Employeenumber: 02342342" -H "unique-id: sms2323@johnshopkins.edu"
```
It should behave the same way with respect to updating user attributes as above. That actually happens in the filter before the services are called.

Verify that the shib user cannot create Grants, but can create Submissions.  The Shib user should be able to modify the created submission, but not delete it. The Shib user can also create publications and create/update File and SubmissionEvent. (The File and SubmissionEvent objects must point to a submission which they are the submitter on.)

How to create a grant. Deal with other objects the same way. Just change the type value.
```
curl   -H "Eppn: sallysubmitter@johnshopkins.edu" -H "Displayname: Sally M. Submitter" -H "Mail: sally232@jhu.edu" -H "Givenname: Sally" -H "Sn: Submitter" -H "Affiliation: FACULTY@johnshopkins.edu" -H "Employeenumber: 02342342" -H "unique-id: sms2323@johnshopkins.edu"  -X POST http://localhost:8080/data/grant -H "accept: application/vnd.api+json" -H "Content-Type: application/vnd.api+json" -d '{"data":{"type":"grant"}}'
```


pass-core-object-service/src/main/java/org/eclipse/pass/object/model/Submission.java
```diff
@@ -38,6 +40,8 @@
  * @author Karen Hanson
  */
 
+@CreatePermission(expression = "User is Backend OR User is Submitter")
```
jabrah, line 43: Will this allow a preparer to create a Submission? My understanding of the Proxy Submission is that the preparer creates the submission and adds relevant data, then the Submitter reviews it and clicks the "submit" button.

The same question applies to the Publication model, since this object would be created with the Submission

jabrah (approved):

jaredgalanis (approved):
Looks good to me. I have not a chance to test it yet, but the headers look like what pass-auth has.

PR #37, opened 2023-04-04
This is the initial port of the 205-java-impl branch of pass-policy-service

jrmartino:
This PR replaces #28 - the only difference is that we have rebased to the new 0.5.0-SNAPSHOT version of main

### DIFF
--- a/pass-core-policy-service/API.md
+++ b/pass-core-policy-service/API.md
@@ -1,0 +1,111 @@
+# Web API
+
+## Policies
+
+The policy service has a `/policies` endpoint that determines the set of policies that are applicable to
+a given submission.  Note:  The results may be dependent on _who_ submits the request.  For example, if
+somebody from JHU invokes the policies endpoint, a general "policy for JHU employees" may be included in the results.
+
+### Policies Request
+
+`GET /policy-service/policies?submission=${SUBMISSION_ID}`
+
+or (with encoded submission=${SUBMISSION_URI}))
+
+```HTTP
+POST /policy-service/policies
+Content-Type application/x-www-form-urlencoded
+```
+
+### Policies Response
+
+The response is a list of IDs to Policy resources, decorated with a `type` property:
+
+```json
+[
+ {
+   "id": "3",
+   "type": "funder"
+ },
+ {
+   "id": "22",
+   "type": "institution"
+ }
+]
+```
+
+## Repositories
+
+The policy service has a `/repositories` endpoint that, for a given submission, calculates the repositories that may be
+deposited into in order to satisfy any applicable policies for that submission.
+
+### Repositories Request
+
+GET `/policy-service/repositories?submission=${SUBMISSION_ID}`
+or, with urlencoded (with encoded submission=${SUBMISSION_ID}) as the body:
+
+```HTTP
+POST /policy-service/repositories
+Content-Type: application/x-www-form-urlencoded
+```
+
+### Repositories Response
+
+Response an application/json document that lists repositories sorted into buckets, as follows:
+
+```json
+{
+   "required": [
+       {
+           "url": "http://pass.local/fcrepo/rest/repositories/1",
+           "selected": true
+       }
+   ],
+   "one-of": [
+       [
+           {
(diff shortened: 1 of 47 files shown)


pass-core-policy-service/API.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+### Repositories Response
+
+Response an application/json document that lists repositories sorted into buckets, as follows:
+
+```json
+{
+   "required": [
+       {
```
markpatton, line 59: Eventually update example response urls.

pass-core-policy-service/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
markpatton, line 1: Not needed here.

pass-core-policy-service/NOTICE.txt
```diff
@@ -0,0 +1,8 @@
+Copyright 2022 Johns Hopkins University
```
markpatton, line 1: Not needed.

pass-core-policy-service/src/main/java/org/eclipse/pass/policy/components/Credentials.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+package org.eclipse.pass.policy.components;
+
+/**
+ * Represents the Credentials object
+ * Contains a client's details
+ *
+ * @author David McIntyre
+ */
+public class Credentials {
```
markpatton, line 24: I don't think this is needed.

pass-core-policy-service/src/main/java/org/eclipse/pass/policy/interfaces/PolicyResolver.java
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Represents PolicyResolver interface
+ * Interpolates any variables in a policy
+ * If policy ID resolves to a list, the list of resolved policies is returned
+ *
+ * @author David McIntyre
+ */
+public interface PolicyResolver {
+
+    public List<PolicyRules> resolve(VariablePinner variables) throws RuntimeException;
```
markpatton, line 32: Use a checked exception for this. Maybe IOException makes sense? Or perhaps a new PolicyServiceException?

pass-core-policy-service/src/main/java/org/eclipse/pass/policy/services/PassPolicyServiceController.java
```diff
@@ -0,0 +1,203 @@
...
+        }
+
+        // call to policy service
+        try {
+            List<RepositoryRules> repositories = policyService.findRepositories(submission, headers);
+        } catch (RuntimeException e) {
+            response.sendError(HttpServletResponse.SC_BAD_REQUEST, e.getMessage());
+        }
+
+        response.getWriter().append("Served at: ").append(request.getContextPath());
```
markpatton, line 179: Looks like a placeholder response.

pass-core-policy-service/src/main/java/org/eclipse/pass/policy/services/PassPolicyServiceController.java
```diff
@@ -0,0 +1,203 @@
...
+            response.getWriter().append(policyResourceArray.toString());
+            LOG.info("Returning policies result for submission " + submission);
+            response.setStatus(200);
+        } catch (RuntimeException e) {
+            LOG.error("Unable to find relevant policies", e);
+            response.sendError(HttpServletResponse.SC_BAD_REQUEST, e.getMessage());
+            return;
+        }
+
+        response.getWriter().append("Served at: ").append(request.getContextPath());
```
markpatton, line 113: This will break the json return value.

markpatton (commented):
This looks like it's coming together.

My biggest comment is just some suggestions about the exception handling. Let's avoid the direct use of RuntimeException where possible. Instead I would just either use IOException where appropriate. Where that doesn't work, just create a blanket PolicyServiceException or even PolicyDSLException.

jrmartino:
closing in favor of a new reimplementation

PR #38, opened 2023-04-13
Update automation publishing Docker image on push to main

jabrah:
* Remove logic dealing with updating hashes or updating the `eclipse-pass.nightly.yml` compose override file
  * This needs to be done in conjunction updates to `pass-docker` which stops using image hashes in compose files
* Updates other parts with community supported actions such as logging into GHCR


markpatton (approved):
Looks good. And test passed.

PR #41, opened 2023-04-19
Update metadata schemas service response

tsande16:
Service should send an array of relevant schemas in its response. If merge=true, return an array with only 1 element, otherwise return individual schemas in an array. 

PR also includes minor documentation fixes.

[Update metadata schemas service response#524](https://github.com/eclipse-pass/main/issues/524)

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ The [file service](pass-core-file-service/README.md) provides a mechanism to per
 
 # Metadata Schema service
 
-The [metadata schema service](pass-core-metadataschema-service/README.md) provides a mechanism to persist files.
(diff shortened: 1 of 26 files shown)


jabrah (changes requested):
Dropped this into the current schema service integration work and found the merged schema I got from the service had some weird looking data that crashed the UI form rendering. Should rescope this change to address what appears to be a bug in the dereferencing and/or merging algorithm of this service.
Provided example responses that can be used in further testing

jabrah (approved):
Looking good. Seems to pass our acceptance tests

jabrah (approved):

PR #42, opened 2023-04-21
initial new implementation of simplified policy service 

jrmartino:
simplified version of the policy service without the DSL logic etc

### DIFF
--- a/pass-core-policy-service/API.md
+++ b/pass-core-policy-service/API.md
@@ -51,11 +51,11 @@ Response an application/json document that lists repositories sorted into bucket
 [
    {
      "url": "http://pass.local/fcrepo/rest/repositories/1",
-     "selected": true
+     "selected": false
    },
    {
      "url": "http://pass.local/fcrepo/rest/repositories/2",
-     "selected": true
+     "selected": false
    },
    {
      "url": "http://pass.local/fcrepo/rest/repositories/3",
@@ -63,7 +63,7 @@ Response an application/json document that lists repositories sorted into bucket
    },
    {
      "url": "http://pass.local/fcrepo/rest/repositories/4",
-     "selected": true
+     "selected":false
    },
    {
      "url": "http://pass.local/fcrepo/rest/repositories/5",
--- a/pass-core-policy-service/API.md
+++ b/pass-core-policy-service/API.md
@@ -1,0 +1,85 @@
+# Web API
+
+## Policies
+
+The policy service has a `/policies` endpoint that determines the set of policies that are applicable to
+a given submission.  Note:  The results may be dependent on _who_ submits the request.  For example, if
+somebody from JHU invokes the policies endpoint, a general "policy for JHU employees" may be included in the results.
+
+### Policies Request
+
+`GET /policy-service/policies?submission=${SUBMISSION_ID}`
+
+### Policies Response
+
+The response is a list of IDs to Policy resources, decorated with a `type` property:
+
+```json
+[
+ {
+   "id": "3",
+   "type": "funder"
+ },
+ {
+   "id": "22",
+   "type": "institution"
+ }
+]
+```
+
+
+## Repositories
+
+The policy service has a `/repositories` endpoint that, for a given submission, calculates the repositories that may be
+deposited into in order to satisfy any applicable policies for that submission.
+
+### Repositories Request
+
+GET `/policy-service/repositories?submission=${SUBMISSION_ID}`
+or, with urlencoded (with encoded submission=${SUBMISSION_ID}) as the body:
+
+```HTTP
+POST /policy-service/repositories
+Content-Type: application/x-www-form-urlencoded
+```
+
+### Repositories Response
+
+Response an application/json document that lists repositories sorted into buckets, as follows:
+
+```json
+[
+   {
+     "url": "http://pass.local/fcrepo/rest/repositories/1",
+     "selected": true
+   },
+   {
+     "url": "http://pass.local/fcrepo/rest/repositories/2",
+     "selected": true
+   },
+   {
+     "url": "http://pass.local/fcrepo/rest/repositories/3",
+     "selected": false
+   },
+   {
+     "url": "http://pass.local/fcrepo/rest/repositories/4",
+     "selected": true
+   },
+   {
+     "url": "http://pass.local/fcrepo/rest/repositories/5",
+     "selected": false
+   },
+   {
+     "url": "http://pass.local/fcrepo/rest/repositories/6",
+     "selected": true
+   }
+]
+
+```
+
+
+Repositories contained within each of the above lists are JSON objects containing the following fields:
+
+* `url`: the URL to the repository resource in Fedora
+* `selected`: optional field.  Specifies if the repository should be selected by default in the UI or not.
+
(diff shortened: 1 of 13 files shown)


jrmartino:
the initial integration smke test gets this error:

org.springframework.web.util.NestedServletException: Request processing failed; nested exception is org.hibernate.LazyInitializationException: failed to lazily initialize a collection of role: org.eclipse.pass.object.model.Policy.repositories, could not initialize proxy - no Session
 basically at at org.eclipse.pass.policy.service.PolicyServiceSimpleImpl.findRepositoriesForSubmission(PolicyServiceSimpleImpl.java:88)

pass-core-policy-service/src/main/java/org/eclipse/pass/policy/service/PolicyServiceSimpleImpl.java
```diff
@@ -0,0 +1,175 @@
...
+            }
+
+            //If the user is an affiliate of the institution, add the institution's policy
+            String user_name = userPrincipal.getName();
+            PassClientSelector<User> userSelector = new PassClientSelector<>(User.class);
+            userSelector.setFilter(RSQL.equals("username", user_name));
+            PassClientResult<User> userResult = passClient.selectObjects(userSelector);
+
+            if (userResult.getObjects().size() == 1
+                && userResult.getObjects().get(0).getAffiliation().contains(institution)
```
jabrah, line 80: Looks fine according to the data model, but I don't think `affiliation` is being set anywhere.

Could update demo data easily to add affiliation info, would also need to update the User Service to add affiliation, as well as anywhere else that could create Users (grant loader? nihms loader?)

pass-core-policy-service/src/main/java/org/eclipse/pass/policy/service/PolicyServiceSimpleImpl.java
```diff
@@ -0,0 +1,175 @@
...
+                && institutionalRepositoryName != null) { //have a unique user in the system
+                PassClientSelector<Repository> repositorySelector = new PassClientSelector<>(Repository.class);
+                repositorySelector.setFilter(RSQL.equals("name", institutionalRepositoryName));
+                PassClientResult<Repository> repositoryResult = passClient.selectObjects(repositorySelector);
+                if (repositoryResult.getObjects().size() == 1) {
+                    repositories.add(repositoryResult.getObjects().get(0));
+                }
+            }
+        }
```
jabrah, line 150: Since you grab the institution's Policy, that should bring the institution's Repository with it. Shouldn't need to do the 2nd institution specific repository search.

It doesn't hurt anything though, since it should be ignored as a duplicate in the Set

pass-core-policy-service/src/main/java/org/eclipse/pass/policy/service/PolicyServiceSimpleImpl.java
jrmartino: I'll look at the user service, I think Affiliations may be set by shib. We don't need the grant loader to set it as it isn't an issue until a user logs in

pass-core-policy-service/API.md
```diff
@@ -0,0 +1,85 @@
...
+   {
+     "url": "http://pass.local/fcrepo/rest/repositories/5",
+     "selected": false
+   },
+   {
+     "url": "http://pass.local/fcrepo/rest/repositories/6",
+     "selected": true
+   }
+]
+
```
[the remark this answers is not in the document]
jrmartino: i can easily change the code to match this behavior - let's talk today

pass-core-policy-service/src/main/java/org/eclipse/pass/policy/service/PolicyServiceSimpleImpl.java
markpatton: The affiliation is set by the user service.

pass-core-policy-service/API.md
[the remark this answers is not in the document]
jrmartino: i'll make this change

pass-core-policy-service/API.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+
+Repositories contained in the above list are JSON objects containing the following fields:
+
+* `url`: the URL to the repository resource in Fedora
```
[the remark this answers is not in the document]
markpatton: Ideally all of the services which return references to objects in the data model so so the same way. But this is something we may have to cleanup later.

 The user service does it as an object {id:4, type:user}. If we want to user a uri, something like type:id would make sense.

pass-core-policy-service/src/main/java/org/eclipse/pass/policy/service/PolicyServiceSimpleImpl.java
```diff
@@ -0,0 +1,175 @@
...
+ * Simple implementation of the Policy Service interface. Provides Sets of policies or repositories
+ *
+ * @author jrm
+ */
+public class PolicyServiceSimpleImpl implements PolicyService {
+
+    private RefreshableElide refreshableElide;
+
+    //defaults for environment variables are for the IT
+    private final String institutionalPolicyTitle = System.getenv("INSTITUTIONAL_POLICY_TITLE") != null ?
```
markpatton, line 46: This configuration should be done using spring properties using "pass.policy" as the prefix. The in application.yaml you can add env variables. The env variables should use the prefix "PASS_CORE_POLICY_".

pass-core-policy-service/src/main/java/org/eclipse/pass/policy/service/PolicyServiceSimpleImpl.java
```diff
@@ -0,0 +1,175 @@
...
+import org.eclipse.pass.object.model.Repository;
+import org.eclipse.pass.object.model.Submission;
+import org.eclipse.pass.object.model.User;
+
+/**
+ * Simple implementation of the Policy Service interface. Provides Sets of policies or repositories
+ *
+ * @author jrm
+ */
+public class PolicyServiceSimpleImpl implements PolicyService {
```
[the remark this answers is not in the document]
jrmartino: ok :-)

pass-core-policy-service/src/main/java/org/eclipse/pass/policy/service/PassPolicyServiceController.java
```diff
@@ -0,0 +1,191 @@
...
+        throws IOException {
+        response.setContentType(MediaType.APPLICATION_JSON_VALUE);
+        response.setCharacterEncoding(StandardCharsets.UTF_8.name());
+
+        LOG.info("Servicing new request......");
+        LOG.debug("Context path: " + request.getContextPath() + "; query string " + request.getQueryString());
+
+        // retrieve submission ID from request
+        String submissionParameter = request.getParameter("submission");
+        Long submissionId = Long.parseLong(submissionParameter);
```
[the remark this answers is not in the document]
jrmartino: done

pass-core-policy-service/src/main/java/org/eclipse/pass/policy/service/PolicyServiceSimpleImpl.java
jrmartino: done

jabrah (approved):

PR #43, opened 2023-04-27
#527 updaet DOI service to use RSQL.hasMember filter to include issns in ElideConnector find() method

jrmartino:
update doi service to use RSQL.hasMember filter for incorporation issns in ElideConnector find() method


markpatton (approved):
Looks good.

PR #45, opened 2023-05-04
Clean code and tests

rpoet-jh:
This PR is for the following issues:

https://github.com/eclipse-pass/main/issues/530
https://github.com/eclipse-pass/main/issues/470

Most changes are fairly straightforward code cleanup as requested in 530.  I used checkstyle, intellij code analysis, and PMD.  I didn't commit the PMD dependency, but let me know if we want to eventually add it.

There are a couple changes that I wanted to highlight:

- The addition of `logback-test.xml` to `pass-core-main/test/resources`.  This sets the default logging level for the console appender for WARN for all tests in this module.  This was added to fix 470.  If you want to see more logging from the tests for debugging purposes, this is the file to change.
- There are slightly more involved changes in the `pass-core-metadataschema-service` module.  I added spring annotations to the classes to clean up the IOC wiring up a bit.  I also changed the unit test mocking in this module to mock at a lower level in elide, this removed some code in the src classes that was there just for testing purposes.

### DIFF
--- a/pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
+++ b/pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
@@ -23,8 +23,6 @@ import org.eclipse.pass.file.service.storage.FileStorageService;
 import org.eclipse.pass.file.service.storage.StorageFile;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.context.annotation.Lazy;
 import org.springframework.core.io.ByteArrayResource;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.MediaType;
@@ -49,14 +47,13 @@ import org.springframework.web.multipart.MultipartFile;
 public class PassFileServiceController {
     private static final Logger LOG = LoggerFactory.getLogger(PassFileServiceController.class);
 
-    @Lazy
-    @Autowired
-    private FileStorageService fileStorageService;
+    private final FileStorageService fileStorageService;
 
     /**
      *   Class constructor.
      */
-    public PassFileServiceController() {
+    public PassFileServiceController(FileStorageService fileStorageService) {
+        this.fileStorageService = fileStorageService;
     }
 
     /**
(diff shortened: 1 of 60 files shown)


pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
```diff
@@ -49,14 +47,13 @@
...
-    @Lazy
-    @Autowired
-    private FileStorageService fileStorageService;
+    private final FileStorageService fileStorageService;
 
     /**
      *   Class constructor.
      */
-    public PassFileServiceController() {
+    public PassFileServiceController(FileStorageService fileStorageService) {
```
tsande16, line 55: Good to know that constructor injection is preferred over annotating a class property.

pass-core-metadataschema-service/src/test/java/org/eclipse/pass/metadataschema/SchemaTestUtils.java
```diff
@@ -0,0 +1,59 @@
...
+import com.yahoo.elide.core.datastore.DataStore;
+import com.yahoo.elide.core.datastore.DataStoreTransaction;
+import com.yahoo.elide.core.dictionary.EntityDictionary;
+
+public final class SchemaTestUtils {
+
+    private SchemaTestUtils() {}
+
+    /**
+     * Returns a RefreshableElideMocked object. The RefreshableElideMocked object contains
```
tsande16, line 20: I like this as a solution for mocking RefreshableElide.

pass-core-object-service/src/test/java/org/eclipse/pass/object/model/FileModelTests.java
```diff
@@ -41,7 +41,7 @@ public void testFileEqualsAndHashCode()  {
         assertEquals(file1.hashCode(), file2.hashCode());
 
         file1.setDescription("different");
-        assertTrue(!file1.equals(file2));
+        assertNotEquals(file1, file2);
```
tsande16, line 44: This is trivial, but I'm assuming this is preferred since it's easier to read?

tsande16 (approved):
Definitely some good code clean up. Learned a few things. Mvn clean install works and unit/ITs pass.

pass-core-object-service/src/test/java/org/eclipse/pass/object/model/FileModelTests.java
rpoet-jh: Yeah, this was a suggestion from IJ.  The assertion method is more succinct since it states the assertion instead of having to derive it from the assertion method and not operator.

markpatton (approved):
Read through pr and ran tests locally. It looks good.

PR #46, opened 2023-05-22
Upgrade to Java 17

rpoet-jh:
Needed files have been updated to Java 17.

Changes are straight forward, but wanted to note that I set the FROM docker image a specific release of Java 17 in the `pass-core-main/Dockerfile` - `eclipse-temurin:17.0.7_7-jre`.  We could also set it to `eclipse-temurin:17-jre`, which would install the latest 17 GA release, but I figured we probably want to control this for releases.  Let me know otherwise.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,10 +18,11 @@ jobs:
     steps:
       - name: "Checkout the repository"
         uses: actions/checkout@v2
-      - name: "Set up JDK 11"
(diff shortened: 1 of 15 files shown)


rpoet-jh:
I also build the docker image for pass-core-main and started pass-docker and ran the acceptance tests.  All passed.

markpatton (approved):
Tried locally and it worked. The cleanups also look good.

PR #47, opened 2023-05-30
Send JMS messages

markpatton:
Changes:
  * Sends JSON messages when some types of objects are created/updated.
  * Connect to either ActiveMQ or Amazon SQS. 
  * Starts an embedded ActiveMQ broker if no connection given.
  * See the documentation in the README
  * Some Spring Boot configuration cleanup
  * Small code cleanups
  * Some pom dependency cleanups.
  * Updated Spring Boot to Elide version.
  * Moved file service config prefix from spring to pass
  * Updated file service aws config to use AWS_REGION for the region and also use the default credentials provider for credentials

To test, run `mvn clean verify`. This will run an integration  against an embedded ActiveMQ

I did some manual testing and was able to connect to SQS and send messages.  In order to test against SQS, use the included docker-compose and edit the .env to enable SQS and set your AWS credentials. Then start it up and verify that creating/updating various objects like Submission with submitted as true emits messages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,9 +40,11 @@ The application is configured by its application.yaml which in turn references a
 By default, pass-core-main, will run with an in memory database. In order to use Postgres, switch to the production profile and set the database environment variables as below.
 Note that the system property `javax.persistence.schema-generation.database.action` can be used to automatically create database tables.
 
-If `PASS_CORE_USE_SQS` is `false` and no connection to a JMS broker is specified with `SPRING_ACTIVEMQ_BROKER_URL`, an embedded ActiveMQ broker will be started.
-If `PASS_CORE_USE_SQS` is `true`, then the connection to Amazon SQS must be configured with `AWS_REGION`, `AWS_ACCESS_KEY_ID`, and `AWS_SECRET_ACCESS_KEY`.
+If `PASS_CORE_USE_SQS` is `true`, then pass-core will attempt to connect to Amazon SQS. The connection must be configured with `AWS_REGION`, `AWS_ACCESS_KEY_ID`, and `AWS_SECRET_ACCESS_KEY`.
 The AWS credentials are also needed if the file service S3 backend is used.
+
+Otherwise a connection to an ActiveMQ broker can be configured by setting `SPRING_ACTIVEMQ_BROKER_URL`. If 'PASS_CORE_EMBED_JMS_BROKER` is true, then an embedded ActiveMQ broker will be started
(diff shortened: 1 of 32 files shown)


README.md
```diff
@@ -40,16 +40,33 @@ The application is configured by its application.yaml which in turn references a
 By default, pass-core-main, will run with an in memory database. In order to use Postgres, switch to the production profile and set the database environment variables as below.
 Note that the system property `javax.persistence.schema-generation.database.action` can be used to automatically create database tables.
 
+If `PASS_CORE_USE_SQS` is `true`, then pass-core will attempt to connect to Amazon SQS. The connection must be configured with `AWS_REGION`, `AWS_ACCESS_KEY_ID`, and `AWS_SECRET_ACCESS_KEY`.
+The AWS credentials are also needed if the file service S3 backend is used.
```
rpoet-jh, line 44: We don't have to figure this out now, but it is best practice when using aws key/secret to use separate keys scoped to the needs of the client.  For example, our SQS client would have a IAM user in AWS with a policy granting access to the SQS queue only.  The File Service would have another IAM user with a policy granting access to the S3 bucket only.

Like I said, we don't have to change this now, but maybe we should create ticket to clean this up later?

rpoet-jh (approved):
Great job, Mark!  I just had one comment, but approving.

pass-core-main/src/main/java/org/eclipse/pass/main/JmsConfiguration.java
```diff
@@ -0,0 +1,233 @@
...
+            return uri.toString();
+        } else {
+            LOG.warn("Cannot create an invitation link, missing required information: " + sub);
+        }
+
+        return null;
+    }
+
+    private String createMessage(SubmissionEvent ev, TokenFactory userTokenFactory) {
+        JsonObjectBuilder ob = Json.createObjectBuilder().add(SUBMISSION_EVENT_KEY, ev.getId().toString()).add(TYPE_KEY,
```
rpoet-jh, line 211: In working on my integration, I realized one thing with the `type` key.  Please set it as a property on the TextMessage.  This is so I can configure the listener to automatically map the text message to a Java object.

So in the `send` method, if you can create the TextMessage and set the `type` as a `textMessage.setStringProperty("type", "SubmissionEvent")` for example, that would help for message listeners.

README.md
markpatton: Created issue to figure this out later: https://github.com/eclipse-pass/main/issues/599

PR #48, opened 2023-06-15
FileService Access Control for Deleting Files

tsande16:
This PR will tighten the access control for deleting files in the `FileService`. Only the `BACKEND` user and the owner of the file can delete the file.

This PR is associated with this issue: https://github.com/eclipse-pass/main/issues/489

There are TODO comments on the File System and S3 integration tests for the check user permissions. This will be refactored in the next ticket to test at the controller level rather than service level. 

- Added method in the Service to check user delete permissions
- Modified the saving of a file to add the user name to the VersionDetails of the OCFL object
- Modified deleteFileById in the controller to check permissions and role of user. BACKEND user has rights to delete all files.

Testing:
`mvn verify`

### DIFF
--- a/pass-core-file-service/pom.xml
+++ b/pass-core-file-service/pom.xml
@@ -91,6 +91,11 @@
             <groupId>jakarta.annotation</groupId>
             <artifactId>jakarta.annotation-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.security</groupId>
+            <artifactId>spring-security-core</artifactId>
+            <version>${spring.security.core.version}</version>
+        </dependency>
     </dependencies>
 
 </project>
--- a/pass-core-file-service/pom.xml
+++ b/pass-core-file-service/pom.xml
@@ -91,11 +91,5 @@
             <groupId>jakarta.annotation</groupId>
             <artifactId>jakarta.annotation-api</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.springframework.security</groupId>
-            <artifactId>spring-security-core</artifactId>
-            <version>${spring.security.core.version}</version>
-        </dependency>
     </dependencies>
-
 </project>
--- a/pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
+++ b/pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
@@ -145,23 +145,25 @@ public class PassFileServiceController {
         ByteArrayResource fileResource;
         String fileId = uuid  + "/" + origFileName;
 
-        //If the file uuid and filename match, but the username does not then return 403.
-
-
-        //check to see if the principal has a file with the UUID and filename. If not then return 403.
+        //Get the file, check that it exists, and get the owner of the file.
         try {
             fileResource = fileStorageService.getFile(fileId);
         } catch (Exception e) {
             LOG.error("File Service: File not found: " + e);
             return ResponseEntity.notFound().build();
         }
-        String userNameInFileName = fileResource.getFilename().split("/")[0];
-        if(!request.isUserInRole(WebSecurityRole.BACKEND.getValue())
-                && !principalName.equals(userNameInFileName)) {
+
+        if(fileResource == null) {
+            return ResponseEntity.notFound().build();
+        }
+
+        if(request.isUserInRole(WebSecurityRole.BACKEND.getValue()) ||
+                fileStorageService.checkUserDeletePermissions(principalName, fileId)) {
+            fileStorageService.deleteFile(fileId);
+            return ResponseEntity.ok().body("Deleted");
+        } else {
             return ResponseEntity.status(HttpStatus.FORBIDDEN)
                     .body("User does not have permission to delete this file.");
         }
-        fileStorageService.deleteFile(fileId);
-        return ResponseEntity.ok().body("Deleted");
     }
 }
--- a/pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
+++ b/pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
@@ -154,12 +154,8 @@ public class PassFileServiceController {
     }
 
     private boolean canUserDeleteFile(String principalName, String fileId, HttpServletRequest request) {
-        if (request.isUserInRole(WebSecurityRole.BACKEND.getValue()) ||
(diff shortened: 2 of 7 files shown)


pass-core-file-service/pom.xml
```diff
@@ -91,6 +91,11 @@
             <groupId>jakarta.annotation</groupId>
             <artifactId>jakarta.annotation-api</artifactId>
         </dependency>
+        <dependency>
```
markpatton, line 94: This spring security dependency is not needed.

pom.xml
```diff
@@ -84,6 +84,7 @@
     <commons.codec.version>1.13</commons.codec.version>
     <spring.boot.version>2.7.5</spring.boot.version>
     <amazon.sqs.version>1.0.4</amazon.sqs.version>
+    <spring.security.core.version>5.7.7</spring.security.core.version>
```
rpoet-jh, line 87: I think this can be removed?

pass-core-main/src/test/java/org/eclipse/pass/file/service/storage/FileStorageServiceS3Test.java
```diff
@@ -68,7 +70,7 @@ void tearDown() {
     void storeFileToS3ThatExists() {
         try {
             StorageFile storageFile = fileStorageService.storeFile(new MockMultipartFile("test", "test.txt",
-                    MediaType.TEXT_PLAIN_VALUE, "Test S3 Pass-core".getBytes()));
+                    MediaType.TEXT_PLAIN_VALUE, "Test S3 Pass-core".getBytes()), USER_NAME);
```
rpoet-jh, line 73: I would recommend verifying in the `ocflRepository` that the associated user of the file is correct.

pass-core-main/src/test/java/org/eclipse/pass/file/service/storage/FileStorageServiceS3Test.java
```diff
@@ -155,4 +157,32 @@ void deleteShouldThrowExceptionFileNotExist() {
...
+
+    //TODO: will be refactored in the next ticket #478
+    @Test
+    void userHasPermissionToDeleteFile() {
+        Boolean hasPermission = false;
+        try {
+            StorageFile storageFile = fileStorageService.storeFile(new MockMultipartFile("test", "test.txt",
+                    MediaType.TEXT_PLAIN_VALUE, "Test Pass-core".getBytes()), USER_NAME);
+            hasPermission = fileStorageService.checkUserDeletePermissions(storageFile.getId(), USER_NAME);
+        } catch (IOException e) {
```
rpoet-jh, line 169: Instead of having try/catch in test, it is recommended to just add IOException to the test method signature.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
```diff
@@ -125,12 +131,36 @@ public ResponseEntity<?> getFileById(@PathVariable String uuid, @PathVariable St
...
+
+        //Get the file, check that it exists, and get the owner of the file.
+        try {
+            fileResource = fileStorageService.getFile(fileId);
+        } catch (Exception e) {
+            LOG.error("File Service: File not found: " + e);
+            return ResponseEntity.notFound().build();
+        }
+
+        if (fileResource == null) {
```
rpoet-jh, line 153: Is this condition possible?

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
```diff
@@ -125,12 +131,36 @@ public ResponseEntity<?> getFileById(@PathVariable String uuid, @PathVariable St
...
+            return ResponseEntity.notFound().build();
+        }
+
+        if (request.isUserInRole(WebSecurityRole.BACKEND.getValue()) ||
```
rpoet-jh, line 157: In accordance with clean code principles, I would suggest the following:

- Put the condition in a method named `canUserDeleteFile`
- Use a terniary for check/call, for example `return canUserDeleteFile(...) ? deleteFile() :  ResponseEntity.status(HttpStatus.FORBIDDEN)...`

rpoet-jh (commented):
Good job, Tim.  I have just a few comments for you to consider.

pom.xml
tsande16: Ah yes. Done.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
tsande16: I don't think so, if a file cannot be found in the `getFile()` it should throw `IOException`. I think this was just me being super defensive. This is now removed.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
rpoet-jh: > Or perhaps you were envisioning something else?

In my example, `deleteFile()` is a new method in `PassFileServiceController`

rpoet-jh: So, something like:

```
private ResponseEntity<?> deleteFile(String fileId) {
    fileStorageService.deleteFile(fileId);
    return ResponseEntity.ok().body("Deleted");
}

pass-core-main/src/test/java/org/eclipse/pass/file/service/storage/FileStorageServiceS3Test.java
tsande16: Done.

pass-core-main/src/test/java/org/eclipse/pass/file/service/storage/FileStorageServiceS3Test.java
tsande16: Done.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
tsande16: Done.

tsande16:
Discussion with Russ: 

- Remove fileOwner on the returned `StoreFile` to the client
- Clean up canUserDeleteFile(), no need for `if` statement

PR #49, opened 2023-06-22
Add Integration Tests to file-service

tsande16:
Add integration tests to test file-service endpoints. Refactor the File System and S3 tests to have a parent class with tests that are shared between the two test classes.

Related Ticket: https://github.com/eclipse-pass/main/issues/478

Notable Changes:

- Split the FileStorageServiceTest into an abstract base class with two child classes: FileStorageServiceFileSystemTest and FileStorageServiceS3Test
- For the S3 tests added `@ActiveProfiles("test-S3")` which has yaml file for its integration test configuration in the `test/resources` folder
- S3 Mock is in a static-initialization block, which is needed to initialize the S3 mocking so that it's running before the application context needs it.
- The AWS credentials are using a `@DynamicPropertySource` method which is required to set the properties before the app context

To test:
`mvn verify` or `mvn clean install`


pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/FileStorageService.java
```diff
@@ -361,7 +363,7 @@ public String getResourceFileRelativePath(String fileId) {
         VersionDetails versionDetails = ocflRepository.describeVersion(ObjectVersionId.head(fileId));
         Collection<FileDetails> allVersionFiles = versionDetails.getFiles();
         Optional<FileDetails> fileDetails = allVersionFiles.stream().findFirst();
-        return fileDetails.get().getStorageRelativePath();
+        return fileDetails.orElse(null).getStorageRelativePath();
```
rpoet-jh, line 366: This would result in a NullPointerEx if the `orElse` executes.  I think you should replace `orElse` with `orElseThrow`.

pass-core-main/src/test/java/org/eclipse/pass/file/service/storage/FileStorageServiceFileSystemTest.java
```diff
@@ -0,0 +1,40 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+
+public class FileStorageServiceFileSystemTest extends FileStorageServiceTest {
+
+    /**
+     * Set up the FileStorageService for testing. Uses the system temp directory for the root directory.
+     */
+    @BeforeEach
+    protected void setUp() {
```
rpoet-jh, line 29: Is empty method needed?

pass-core-main/src/test/java/org/eclipse/pass/file/service/storage/FileStorageServiceS3Test.java
```diff
@@ -1,167 +1,73 @@
...
-                        MediaType.TEXT_PLAIN_VALUE, "Test Pass-core".getBytes()), USER_NAME);
-                assertFalse(fileStorageService.getResourceFileRelativePath(storageFile.getId()).isEmpty());
-            } catch (IOException e) {
-                assertEquals("An exception was thrown in storeFileWithDifferentLangFilesNames. On charset=" + k,
-                        e.getMessage());
-            }
-        });
+    @BeforeEach
+    protected void setUp() {
+        if (!s3MockStarted) {
```
rpoet-jh, line 57: I wonder if the `s3MockApi` can be asked if it is started rather than tracking with the `s3MockStarted` boolean?

pass-core-main/src/test/java/org/eclipse/pass/file/service/storage/FileStorageServiceTest.java
```diff
@@ -124,34 +153,170 @@ void storeFileWithDifferentLangFilesNames() {
...
     }
 
-    //TODO: will be refactored in the next ticket #478
+    /**
+     * Store file, and then check user permissions on that file. User does not have permissions to delete file.
+     *
+     * @throws IOException if there is an error
+     */
     @Test
     void userNoPermissionToDeleteFile() throws IOException {
```
rpoet-jh, line 185: Ideally, this test should be changed to be a true integration test at the Controller level, and verify the expected response code.  Probably same with `deleteShouldThrowExceptionFileNotExist`.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/FileStorageService.java
```diff
@@ -372,15 +374,14 @@ public String getResourceFileRelativePath(String fileId) {
...
      */
     public String getFileContentType(String fileId) {
-        VersionDetails versionDetails = ocflRepository.describeVersion(ObjectVersionId.head(fileId));
-        FileDetails fileDetails = versionDetails.getFiles().stream().findFirst().get();
-        Path fileDetailPath = Paths.get(fileDetails.getPath());
-        File file = fileDetailPath.toFile();
-        //get the content type from the file
         try {
+            VersionDetails versionDetails = ocflRepository.describeVersion(ObjectVersionId.head(fileId));
+            FileDetails fileDetails = versionDetails.getFiles().stream().findFirst().orElse(null);
```
rpoet-jh, line 379: Same NPE concern here.

pass-core-main/src/test/java/org/eclipse/pass/file/service/storage/FileStorageServiceFileSystemTest.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    /**
+     * Cleanup the FileStorageService after testing. Deletes the root directory.
+     */
+    @AfterEach
+    @Override
+    protected void tearDown() throws IOException {
```
rpoet-jh, line 37: I don't think this method is needed either since super will be executed if not overriden here.

rpoet-jh (commented):
Nice work, Tim!  I just have a few comments.

pass-core-main/src/test/java/org/eclipse/pass/file/service/storage/FileStorageServiceS3Test.java
tsande16: So I looked into this one and I don't see anything that indicates whether it's a run state. I could probably check to see if the logger has been instantiated, but I wasn't sure about that since it's not a guarantee that it would running.

rpoet-jh: You can just call stop and start in the setUp method, like:

```
s3MockApi.stop();
s3MockApi.start();

File: pass-core-main/src/test/java/org/eclipse/pass/file/service/storage/FileStorageServiceTest.java (hunk 4 above)
Comment (tsande16) [reply pass-core-main/src/test/java/org/eclipse/pass/file/service/storage/FileStorageServiceTest.java:185]: I attempted this for a second, but realized I'm not sure if we can do this, unless we added a new test user for the credentials. The back-end user authorization is required for all the requests. 

Well I suppose this would depend, if I test canUserDeleteFile() in the controller or if I test deleteFileById(). It sounds like you are suggesting to test the actually DELETE method. If so, do you have a suggestion for working around the authorization? Should I create another test user?

Comment (rpoet-jh) [reply pass-core-main/src/test/java/org/eclipse/pass/file/service/storage/FileStorageServiceTest.java:185]: We will need authorization tests like this going forward, so I think it is worth it to add another user that can be authenticated but not authorized to delete the file associated with another user.  Not sure, but maybe adding another user is as easy as adding another `user` in the yaml here: https://github.com/eclipse-pass/pass-core/blob/8da9b89e2b63353fcb0e3ad17e607b549082e1db/pass-core-main/src/main/resources/application.yaml#L65

File: pass-core-main/src/test/java/org/eclipse/pass/file/service/storage/FileStorageServiceS3Test.java (hunk 3 above)
Comment (rpoet-jh) [reply pass-core-main/src/test/java/org/eclipse/pass/file/service/storage/FileStorageServiceS3Test.java:57]: Any time you can remove a conditional on a boolean flag, it is worth it imo.  I know it is just a test, but it decreases code complexity by removing the conditional.

Comment (tsande16) [reply pass-core-main/src/test/java/org/eclipse/pass/file/service/storage/FileStorageServiceS3Test.java:57]: That's true, get too many boolean variables floating around, can be cumbersome to keep track of what's going on.

Comment (tsande16) [reply pass-core-main/src/test/java/org/eclipse/pass/file/service/storage/FileStorageServiceS3Test.java:57]: Done.

File: pass-core-main/src/test/java/org/eclipse/pass/file/service/storage/FileStorageServiceFileSystemTest.java (hunk 2 above)
Comment (tsande16) [reply pass-core-main/src/test/java/org/eclipse/pass/file/service/storage/FileStorageServiceFileSystemTest.java:29]: Done. There was also an issue with the clean up of temp files that I needed to resolve as well.

File: pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/FileStorageService.java (hunk 1 above)
Comment (tsande16) [reply pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/FileStorageService.java:366]: Done.

File: pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/FileStorageService.java (hunk 5 above)
Comment (tsande16) [reply pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/FileStorageService.java:379]: Done.

File: pass-core-main/src/test/java/org/eclipse/pass/file/service/storage/FileStorageServiceFileSystemTest.java (hunk 6 above)
Comment (tsande16) [reply pass-core-main/src/test/java/org/eclipse/pass/file/service/storage/FileStorageServiceFileSystemTest.java:37]: Done.

File: pass-core-main/src/test/java/org/eclipse/pass/file/service/storage/FileStorageServiceTest.java (hunk 4 above)
Comment (tsande16) [reply pass-core-main/src/test/java/org/eclipse/pass/file/service/storage/FileStorageServiceTest.java:185]: From our slack conversation, we decided that it was best to break this out into its own ticket since their is more involvement with authentication: https://github.com/eclipse-pass/main/issues/645

File: pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/FileStorageService.java [hunk 7]
```diff
@@ -361,7 +366,7 @@ public String getResourceFileRelativePath(String fileId) {
         VersionDetails versionDetails = ocflRepository.describeVersion(ObjectVersionId.head(fileId));
         Collection<FileDetails> allVersionFiles = versionDetails.getFiles();
         Optional<FileDetails> fileDetails = allVersionFiles.stream().findFirst();
-        return fileDetails.get().getStorageRelativePath();
+        return fileDetails.orElseThrow().getStorageRelativePath();
```
rpoet-jh, line 369: Preferred used is this method with a Supplier: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Optional.html#orElseThrow(java.util.function.Supplier)

Throwing a more descriptive runtime exception with message.

rpoet-jh: Style comment:  I find the code more readable like:

```
FileDetails fileDetails = allVersionFiles.stream().findFirst()
    .orElseThrow(() -> SomeEx("Something wrong");
return fileDetails.getStorageRelativePath();

File: pass-core-main/src/test/java/org/eclipse/pass/file/service/storage/FileStorageServiceFileSystemTest.java [hunk 8]
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.eclipse.pass.file.service.storage;
+
+/**
+ * Application Context sets up the FileService. The default of the FileService is FILE_SYSTEM, therefore no
+ * setup is needed here.
+ */
+public class FileStorageServiceFileSystemTest extends FileStorageServiceTest {
```
rpoet-jh, line 22: Instead of the empty class, could you just make `FileStorageServiceTest` not abstract and be the test for FILE_SYSTEM?  `FileStorageServiceS3Test` can still `FileStorageServiceTest` as is pretty sure.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/storage/FileStorageService.java
tsande16: Done.

pass-core-main/src/test/java/org/eclipse/pass/file/service/storage/FileStorageServiceFileSystemTest.java
tsande16: Done.

rpoet-jh (approved):
Good job, Tim, approved!

PR #50, opened 2023-07-18
Allow AWS endpoint override

rpoet-jh:
This PR makes it possible to override the AWS endpoint on the client library.  It is needed for local testing.

The change is part of deposit-services work: https://github.com/eclipse-pass/main/issues/580


markpatton (approved):
Code makes sense. I am able to run the tests locally.

PR #51, opened 2023-08-01
Fix a couple bugs in file service

markpatton:
Changes:
  * Handle the case of Files.probeContentType returning null. This prevented some files from working.
  * Handle the case of larger files by increasing the limit to 100 MB as specified by pass-ui.


markpatton:
Ok. There is also a bug if there is no file extension. It makes the S3 code unhappy.

tsande16:
@markpatton the issue should now be resolved.

tsande16 (approved):
Looks good!

PR #52, opened 2023-08-11
Fix updates to existing user on login

markpatton:
The check against information provided by the shib request against an existing user object made too many assumptions. This pr fixes the equality checks and add a test.


rpoet-jh (approved):
Good job, Mark.

tsande16 (approved):
Looks good!

PR #53, opened 2023-08-22
Create indexes on User locator ids and Journal issns

markpatton:
User locator ids and Journal issns are both lists of string that have to be indexed appropriately in order for searches on them to be fast. This pr adds indexes on the tables which handle those properties.

### DIFF
--- a/pass-core-object-service/src/main/java/org/eclipse/pass/object/model/Journal.java
+++ b/pass-core-object-service/src/main/java/org/eclipse/pass/object/model/Journal.java
@@ -18,10 +18,12 @@ package org.eclipse.pass.object.model;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
+import javax.persistence.CollectionTable;
 import javax.persistence.ElementCollection;
 import javax.persistence.Entity;
 import javax.persistence.EnumType;
 import javax.persistence.Enumerated;
+import javax.persistence.Index;
 import javax.persistence.ManyToOne;
 import javax.persistence.Table;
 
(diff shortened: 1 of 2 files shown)


rpoet-jh (commented):
Good job, Mark.  The indexes make sense.  However, thinking about how the DDL gets applied to the database, I think it may be better to write a SQL script to make these changes, then we run it in stage manually for this PR.  This would also result in removing the `javax.persistence.Index` annotations, indexes would be recorded in liquibase changesets.  We can then move quickly to put liquibase in place to take care of future DDL changes.  This way, we don't have to wipe out the database to make apply these changes.

markpatton:
Upgrade script for the database. I tested against a database  generated from previous version of pass-core and then was able to successfully use this version of pass-core. Also removed the index annotations so it just does the table name change.

```
ALTER TABLE journal_issns RENAME TO pass_journal_issns;
ALTER TABLE user_locatorids RENAME TO pass_user_locatorids;

CREATE INDEX pass_journal_issns_id_ix ON pass_journal_issns (journal_id);
CREATE INDEX pass_journal_issns_issn_ix ON pass_journal_issns (issns);

CREATE INDEX pass_user_locatorids_id_ix ON pass_user_locatorids (user_id);
CREATE INDEX pass_user_locatorids_locator_ix ON pass_user_locatorids (locatorids);
```

rpoet-jh (approved):
Looks good, Mark!

rpoet-jh:
> ALTER TABLE user_locatorids RENAME TO pass_user_locatorids;

One change to executed script for the statement above, it changed to:

```
ALTER TABLE pass_user_locatorids RENAME TO pass_user_locators;
```

To match name in java

PR #56, opened 2023-08-29
Enable hasmember searching on RepositoryCopy.externalids

markpatton:
- Enables hasmember on Repository.externalids in model
- Add table for RepositoryCopy.externalids
- Fix RepositoryCopy.externalids equality testing
- Remove unused converters


pass-core-main/src/main/resources/db/changelog/changelog.yaml
```diff
@@ -12,4 +12,18 @@ databaseChangeLog:
              encoding: utf-8
              path: /db/changelog/schema/initial-schema.sql
              splitStatements: true
-             stripComments: true
\ No newline at end of file
+             stripComments: true
+  -  changeSet:
+       id:  2
+       author:  mark-patton
+       preConditions:
```
rpoet-jh, line 19: This isn't needed.  The preCondition was only needed on the initial schema changeset because it was executed against a database that already had all the DDL.  All subsequent changesets shouldn't have it.

tsande16 (approved):
Looks good! Ran clean install and builds successfully. Ran a smoke test creating a RepositoryCopy and then querying it using the `externalId`, which resulted in the `RepositoryCopy` being returned.

pass-core-main/src/main/resources/db/changelog/changelog.yaml
markpatton: Got it. Will adjust.

PR #57, opened 2023-08-29
Add repository_copy externalids data migration script

rpoet-jh:
https://github.com/eclipse-pass/main/issues/702


markpatton (approved):
Makes sense. Runs for me locally.

PR #58, opened 2023-08-29
Add indexes for common queries

markpatton:
Indexes added for most common queries.
For award number and journal name there are postgres specific indexes.

In order to test, try running pass-core by itself using the builtin db and also with postgres.

### DIFF
--- a/pass-core-main/src/main/resources/db/changelog/changelog.yaml
+++ b/pass-core-main/src/main/resources/db/changelog/changelog.yaml
@@ -36,3 +36,26 @@ databaseChangeLog:
             columns:
             - column:
                 name: externalids
+  -  changeSet:
+       id:  4
+       author:  mark-patton
+       changes:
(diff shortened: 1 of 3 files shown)


pass-core-main/src/main/resources/db/changelog/changelog.yaml
```diff
@@ -36,3 +36,26 @@ databaseChangeLog:
...
+       changes:
+         - sqlFile:
+             encoding: utf-8
+             path: /db/changelog/schema/indices.sql
+             splitStatements: true
+             stripComments: true
+  -  changeSet:
+       id:  5
+       author:  mark-patton
+       preConditions:
```
rpoet-jh, line 51: Should replace this `preCondition` with `dbms: postgresql`

rpoet-jh (approved):
Good job, Mark!

PR #59, opened 2023-09-12
File Service - Update Documenation and Verify Temp Storage

tsande16:
- README needs to make clear the role of PASS_CORE_FILE_SERVICE_ROOT_DIR when configured to use S3.
- In staging the name of the directory should default to the OS generated name and not the name of the variable itself.
- Verify that the temp storage is being cleaned up when configured using S3
- Verify that if PASS_CORE_FILE_SERVICE_ROOT_DIR is configured with a bucket name that it continues to work correctly using the temp directories. NOTE: This is no longer valid. PASS_CORE_FILE_SERVICE_ROOT_DIR must be a valid path on the local file system or attached file/network storage that is accessible by a path on the OS.

Things to note:

- Recognition of the behavior of the FileService and how `PASS_CORE_FILE_SERVICE_ROOT_DIR` is supposed to be used is noted in the README
- New test was created to test that the temp directory is cleaned up after a file is uploaded.

To Test:
`Mvn verify`

**To check that files are uploaded, they are logged in the logs:
Example:**
`INFO  o.e.p.f.s.storage.FileStorageService.storeFile - File Service: File with ID 469d149c-df88-4baa-b7a2-793ba6d91ce0/Nanometer-Scale Thermometry.pdf was stored in the file system repo at the location:C:\Users\tsande16\AppData\Local\Temp\1944912672741001035\ocfl\211\b3c\73a\211b3c73a7aaf393803607b3be9710a16169e89375a0eac1648feff7e4aa9173\v1\content\469d149c-df88-4baa-b7a2-793ba6d91ce0.pdf`

**To check your temp, the location is logged here:**
`14-09-2023 11:38:49.362 [main] [pass-core, ] INFO  o.e.p.f.s.storage.FileStorageService.<init> - File Service: C:\Users\tsande16\AppData\Local\Temp\1944912672741001035 Storage Root Directory`

I also tested locally running the jar and checking my local file system to ensure that the temp directory was being cleaned up after uploading files. I also tested in docker by running `docker-compose up -d` and uploading files there and checking the `/tmp` directory to ensure that was being cleaned up as well. Also to ensure that `PASS_CORE_FILE_SERVICE_ROOT_DIR` wasn't being created as a temp directory I grepped the filesystem in docker.


markpatton (approved):
The changes look good and tests pass for me locally.

PR #60, opened 2023-09-12
[pass-core] Submission and Deposit entity optimistic locking

rpoet-jh:
https://github.com/eclipse-pass/main/issues/715

This adds the `@Version` version attribute to the Submission and Deposit entities to enable optimistic locking on updates.  Incrementing and setting the version values is handled by JPA/hibernate.

The elide lifecycle code is needed to handle the `Lost Update` scenario covered in the `ElidePassClientTest`.  This code is needed because of the way elide saves updates in the Patch handler.  It does not call `em.merge` because is sees the entities as managed.  So I put in the lifecycle hooks on updates for Submission and Deposit before flush to check the version of the request entity against what is in the database and handle appropriately.

The Elide configuration was also changed to return `200` for updates instead of `204`.  By doing this, the updated entity json is returned in the response.  This is important for versioned entities like Submission and Deposit so that the new version value can be set in the client's object from the response.

There will be another PR opened in pass-support for the pass-data-client to add the version attributes and to handle CONFLICT errors as needed, but I figured I'd open this for review now to get this solution checked.

https://github.com/eclipse-pass/pass-support/pull/57

And this PR in pass-ui:

https://github.com/eclipse-pass/pass-ui/pull/1209

### DIFF
--- a/pass-core-main/src/main/java/org/eclipse/pass/main/JmsConfiguration.java
+++ b/pass-core-main/src/main/java/org/eclipse/pass/main/JmsConfiguration.java
@@ -19,6 +19,7 @@ import java.net.URI;
 import java.util.HashMap;
 import java.util.Objects;
 import java.util.Set;
+import java.util.TimeZone;
 import javax.jms.ConnectionFactory;
 import javax.jms.TextMessage;
 import javax.json.Json;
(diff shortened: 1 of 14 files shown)


rpoet-jh:
I did some testing with pass-docker with this PR and the pass-support.  There are pass-ui changes needed as well, when it reads a submission/deposit and saves, the version value must be passed in the patch payload.  I'll check with @jaredgalanis on what is needed to make this change.

rpoet-jh:
> I did some testing with pass-docker with this PR and the pass-support. There are pass-ui changes needed as well, when it reads a submission/deposit and saves, the version value must be passed in the patch payload. I'll check with @jaredgalanis on what is needed to make this change.

 I opened this PR for the UI change: https://github.com/eclipse-pass/pass-ui/pull/1209

rpoet-jh:
> I see there are different locking tests for ElidePassClient and ElideDataStorePassClient. Is there some behavior difference between the two client implementations?

Yeah, the `ElideDataStorePassClient` invokes methods on the Elide transaction directly, the `ElidePassClient` uses the `Elide` class to perform CRUD operations.  What this means for optimistic locking is the `ElideDataStorePassClient` triggers the checks for covering Lost Update which happens in the entity manager merge method, but `ElidePassClient` doesn't as described in the PR desc.

PR #61, opened 2023-09-14
Update the database for migration

markpatton:
The colums SubmissionEvent.comment, File.description, and Publication.title need to support larger values. This pr changes those columns to text.

In addition Funder.localKey and Grant.localKey are indexed to support the grant loader which does queries on those values.


rpoet-jh (approved):
Good job, Mark.

PR #62, opened 2023-09-18
Remove Contributor and Publisher from the data model

markpatton:
Removes unused objects from the data model.


tsande16 (approved):
Looks good! Mvn clean install runs successfully for me locally.

PR #64, opened 2023-09-29
Modify the repositories policy service to respect effective policies

markpatton:
Modify the policy repositories service to return repositories from the intersection of the effective and computed policies of the submission.

This seems to be what the old policy service did. It should allow the don't deposit to PMC checkbox to work.


tsande16 (approved):
mvn clean install runs locally. Looks good!

PR #65, opened 2023-10-05
Make journal-title optional for JScholarship and required for NIHMS

markpatton:
This pr removes journal-title from the required list in global.json since it is optional for JScholarship. The individual repositories now must make it required when needed. In addition fields that are directly derived from Journal objects are marked as read only. The manuscript title is also marked read only since it is entered in the first step. In addition the `(required)` in labels is removed on read only fields. The toolbar and buttons for the read only issns are disabled.


jaredgalanis (approved):
Looks good to me 👍

PR #67, opened 2023-10-09
Change default value from #{null} to empty string

rpoet-jh:
In file-service configuration.

https://github.com/eclipse-pass/main/issues/740


tsande16 (approved):
`mvn clean install` runs successfully locally. Looks good!

PR #68, opened 2023-10-16
Release workflow updates

jabrah:
Resolves https://github.com/eclipse-pass/main/issues/752

* Prevent release automation from pushing commits early
  * Workflow will need to make it through the whole maven release plugin step before any commits are pushed up to GH
* Remove bad argument from maven command
  * This arg was an attempt to quiet some Maven outputs, but proved to have no effect


markpatton (approved):
Looks good.

PR #71, opened 2023-11-03
Add institutional liquibase changelog

rpoet-jh:
Now the db.changelog includes a public changelog and an institutional changelog.  The public changelog will be all changes that we want to release to the community.  The institutional changelog will contain changes specific to the institution that has deployed pass.

In JHU case, I will add a db.changelog to the pass-jhu-aws repo, and modify the deployment to include this in the mounted volume of pass-core.  I will do this once this PR is approved and merged.


markpatton (approved):
Makes sense. I tested locally and it looks good.

PR #73, opened 2023-12-13
Refactor build and release workflows

markpatton:
Switch to reusable snapshot and release workflows from main. Note that there is no longer a separate update image workflow. For simplicity, the snapshot workflow also updates the images.

Testing resulted in two full releases being made, 1.3.0-RC1 and 1.3.0-RC2. The commits for RC1 went to a branch and were lost when history was rewritten. The commits for RC2 went to main.  RC1 and RC2 should be identical in any case. After merge the RC1 tag should be updated to be the same as the RC2 tag.


jabrah (approved):
Looks good to me, especially given the test runs

PR #74, opened 2024-01-03
Add Deposit.statusMessage attribute

rpoet-jh:
Needed as part of nihms email integration.  The main PR will be opened in pass-support, but this change is needed for ITs to pass in pass-support.

https://github.com/eclipse-pass/main/issues/856


markpatton (approved):
Looks good.

PR #75, opened 2024-01-24
Set logger to warn for aws sqs logging

rpoet-jh:
To remove sqs info logs in prod log.


markpatton (approved):
Yay for fewer log messages!

PR #76, opened 2024-01-24
874 file deletion bug fix

tsande16:
There was a bug in the file service which was not deleting a file for an owner of a file. The ITs still passed because the ITs use BACKEND as the user. This user has permission to delete all files and therefore did not reach the point of the exception that was being thrown. 

Notable Changes:

- Fix for the bug
- Added exceptions to better handle when an error occurs on file deletion and permission checks.

Notes: Discussed and went over the changes with @jabrah. I needed to rebuild pass-ui in order to smoke test, but John said he would check this for me when reviewing this PR.

### DIFF
--- a/pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
+++ b/pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
@@ -154,12 +154,22 @@ public class PassFileServiceController {
     }
 
     private boolean canUserDeleteFile(String principalName, String fileId, HttpServletRequest request) {
-        return (request.isUserInRole(WebSecurityRole.BACKEND.getValue()) ||
-                fileStorageService.checkUserDeletePermissions(principalName, fileId));
+        try {
+            return (request.isUserInRole(WebSecurityRole.BACKEND.getValue()) ||
+                    fileStorageService.checkUserDeletePermissions(fileId, principalName));
+        } catch (Exception e) {
+            LOG.error("File Service: Unable to determine user permissions to delete file: " + e);
+            return false;
+        }
     }
 
     private ResponseEntity<?> deleteFile(String fileId) {
-        fileStorageService.deleteFile(fileId);
-        return ResponseEntity.ok().body("Deleted");
+        try {
+            fileStorageService.deleteFile(fileId);
+            return ResponseEntity.ok().body("Deleted");
+        } catch (Exception e) {
+            LOG.error("File Service: Unable to delete file: " + e);
+            return ResponseEntity.notFound().build();
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
```diff
@@ -154,12 +154,22 @@ public ResponseEntity<?> deleteFileById(@PathVariable String uuid, @PathVariable
...
     }
 
     private ResponseEntity<?> deleteFile(String fileId) {
-        fileStorageService.deleteFile(fileId);
-        return ResponseEntity.ok().body("Deleted");
+        try {
+            fileStorageService.deleteFile(fileId);
+            return ResponseEntity.ok().body("Deleted");
+        } catch (Exception e) {
+            LOG.error("File Service: Unable to delete file: " + e);
```
rpoet-jh, line 171: There should be a test to cover this.  I know there is an open issue for security role tests, but mocking could be used throw the exception from the appropriate method in `canUserDeleteFile`

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
```diff
@@ -154,12 +154,22 @@ public ResponseEntity<?> deleteFileById(@PathVariable String uuid, @PathVariable
...
+            return ResponseEntity.notFound().build();
```
rpoet-jh, line 172: I think 401 would be more appropriate if unauthorized is the only exception being caught, if it is more generic, it should be 400.

rpoet-jh (commented):
Good job, Tim.  Couple comments.

markpatton (approved):
Tried it out locally and the fix looks good. I agree with @rpoet-jh that a test to cover this case would be good.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
tsande16: Done.

pass-core-file-service/src/main/java/org/eclipse/pass/file/service/PassFileServiceController.java
tsande16: Done.

tsande16:
There is a strange issue that is occurring with maven and the application context. @rpoet-jh Found a work-around and it's fixed with commit [0ed44fd](https://github.com/eclipse-pass/pass-core/pull/76/commits/0ed44fd65b88f2668d76da654780bc816d006e6b). @jabrah You may want to pull this latest and do your sanity tests again to ensure everything is fine.

jabrah (approved):
Local testing against a corresponding pass-ui branch seemed to go through fine

PR #77, opened 2024-01-25
Add delete permissions on File models

jabrah:
Part of https://github.com/eclipse-pass/main/issues/514

Would we want to first check if the uploaded bytes on the file system are gone (referenced in `file.uri`)?


jabrah:
IMO it's a little awkward, but manageable for the UI to manage the file bytes and File entity in this way. I agree, though, that we should wait to see if there's feedback about file it before looking into potential changes

markpatton (approved):
Tests pass for me locally. Looks good.

PR #80, opened 2024-02-14
Upgrade elide/spring-boot/aws

rpoet-jh:
The following is in this PR:

- Upgrade elide to 7.x
- Upgrade spring-boot to 3.x
- Upgrade amazon sqs to 2.x
- Updated activemq to activemq artemis
- Updated spring security to 6.x api
- Cleaned metadata tests

There will also be small PRs in pass-support and pass-docker that need to go with this one.


markpatton (approved):
Looks great. This is a big improvement. I did a bunch of local testing and there were no problems.

Only minor issue I noticed is pass.jms.embed being set for two sets even though it has been superseded: ./pass-core-main/src/test/java/org/eclipse/pass/main/JmsSqsConfigurationTest.java and ./pass-core-main/src/test/java/org/eclipse/pass/main/JmsSqsEndpointConfigurationTest.java

PR #81, opened 2024-02-22
Add Delete permissions for Submission and Publication

jabrah:
## Question:

Are we OK with having a multi-step process for deleting and cleaning up a draft Submission and related entities? Deleting draft submissions does not produce SubmissionEvents (from the UI) so shouldn't trigger notification services.

To deleted a submission, a client would need to:
- Remove associated Files, since these Files reference the Submission)
- Remove the submission
- I think removing the submission's Publication, would be last, since the submission would hold a reference to it ?
  - only if no other submissions reference the Publication

There are multiple points of failure in this chain, none of which are really recoverable by the client, although chances of errors should be pretty low.

### DIFF
--- a/pass-core-main/src/test/java/org/eclipse/pass/main/security/AccessControlTest.java
+++ b/pass-core-main/src/test/java/org/eclipse/pass/main/security/AccessControlTest.java
@@ -327,7 +327,7 @@ public class AccessControlTest extends ShibIntegrationTest {
 
             Response response = client.newCall(request).execute();
 
-            check(response, 403);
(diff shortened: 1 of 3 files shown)


markpatton:
You will have to remove the Publication first, and then the Files, and then the Submission. Removing the submission or publication will fail if other things point to them, but this should work fine for draft submissions.

markpatton (approved):
Looks good.

jabrah:
Tested locally fine after some UI tweaks to fix some bad rsql queries

PR #84, opened 2024-04-23
Make the SAML login processing URL configurable.

markpatton:
The SAML login processing URL must be manually configured to match the ACS URL sent to the IDP.


rpoet-jh (approved):
Good job, Mark!

PR #86, opened 2024-04-25
Set SAML ForceAuthn to true in order to require IDP login

markpatton:


rpoet-jh (approved):
Good job, Mark.  On a side note, how lucky to have this exact scenario in the spring sec [docs](https://docs.spring.io/spring-security/reference/servlet/saml2/login/authentication-requests.html#servlet-saml2login-sp-initiated-factory-custom-authnrequest) 😀

PR #87, opened 2024-05-16
Ignore pushes to main with tags for snapshot workflow

rpoet-jh:
Needed for complete workflow PR: https://github.com/eclipse-pass/main/pull/980


markpatton (approved):
Looks good.

PR #88, opened 2024-05-16
Add metadata schema for InvenioRDM along with a test

markpatton:
As a first step for InvenioRDM suppport can just use a simple metadata schema that requires and author and publication date.

Later we may want to get more detailed name info from authors later on and so some work in pass-ui to get that work from the DOI.

Documented findings here: https://docs.google.com/document/d/1JllSJ48d14AvIVyhB0l0zUTcoj1vSB2oNlKRwi2wAVs/edit?usp=sharing


tsande16 (approved):
Looks good! Added tests pass locally for me. The only thing I might add, is adding some different schemas to test the robustness of the metadata schema service, but since this was a simple test the scope is fine. Really won't know how much Caltech metadata deviates from our global.json until we start developing the integration.

PR #90, opened 2024-05-22
Turn on CSRF protection. 

markpatton:
Turns on CSRF protection using cookies and headers. Responses from the API will set a cookie XSRF-TOKEN and non-GET requests from the client must set the header X-XSRF-TOKEN with the value of the cookie. GET requests to the /doi service are also protected because they have side effects.

In addition the integration tests have been refactored to handle the CSRF token and do a little more testing.

By default CSRF protection requires a POST to to /logout. This is disabled for now because the a POST to that endpoint with the header set returns a login form with a CSRF token hidden in the form. Submitting that form does do the logout. The format of the token in the in the form is different from the cookie. In addition doing a POST to /logout will initiate the SAML logout process. The complications don't seem to make protecting /logout worthwhile at the moment.

Clients can user any CSRF token they want. The cookie and header just must be consistent. 

Required prs:
  *  https://github.com/eclipse-pass/pass-support/pull/116
  * https://github.com/eclipse-pass/pass-ui/pull/1276
  * https://github.com/eclipse-pass/pass-docker/pull/375

Acceptance tests pass, but require the updated pass-core and pass-support images running in the updated pass-docker.

### DIFF
--- a/pass-core-main/src/main/java/org/eclipse/pass/main/security/CsrfCookieFilter.java
+++ b/pass-core-main/src/main/java/org/eclipse/pass/main/security/CsrfCookieFilter.java
@@ -9,6 +9,9 @@ import jakarta.servlet.http.HttpServletResponse;
 import org.springframework.security.web.csrf.CsrfToken;
 import org.springframework.web.filter.OncePerRequestFilter;
 
+/**
+ * Ensure CSRF token cookie is added.
+ */
 public class CsrfCookieFilter extends OncePerRequestFilter {
     @Override
     protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
--- a/pass-core-main/src/main/java/org/eclipse/pass/main/security/CsrfCookieFilter.java
+++ b/pass-core-main/src/main/java/org/eclipse/pass/main/security/CsrfCookieFilter.java
@@ -1,0 +1,22 @@
+package org.eclipse.pass.main.security;
+
+import java.io.IOException;
+
+import jakarta.servlet.FilterChain;
+import jakarta.servlet.ServletException;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
+import org.springframework.security.web.csrf.CsrfToken;
+import org.springframework.web.filter.OncePerRequestFilter;
+
+public class CsrfCookieFilter extends OncePerRequestFilter {
+    @Override
+    protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
+            throws ServletException, IOException {
+        CsrfToken csrfToken = (CsrfToken) request.getAttribute("_csrf");
+        // Render the token value to a cookie by causing the deferred token to be loaded
+        csrfToken.getToken();
+
+        filterChain.doFilter(request, response);
+    }
+}
--- a/pass-core-main/src/main/java/org/eclipse/pass/main/security/CsrfCookieFilter.java
+++ b/pass-core-main/src/main/java/org/eclipse/pass/main/security/CsrfCookieFilter.java
@@ -10,7 +10,8 @@ import org.springframework.security.web.csrf.CsrfToken;
 import org.springframework.web.filter.OncePerRequestFilter;
 
 /**
- * Ensure CSRF token cookie is added.
+ * Ensure CSRF token cookie is added. See
+ * https://docs.spring.io/spring-security/reference/servlet/exploits/csrf.html#csrf-integration-javascript-spa
  */
 public class CsrfCookieFilter extends OncePerRequestFilter {
     @Override
--- a/pass-core-main/src/main/java/org/eclipse/pass/main/security/SecurityConfiguration.java
+++ b/pass-core-main/src/main/java/org/eclipse/pass/main/security/SecurityConfiguration.java
@@ -37,10 +37,12 @@ import org.springframework.security.saml2.provider.service.web.authentication.Sa
 import org.springframework.security.web.SecurityFilterChain;
 import org.springframework.security.web.authentication.logout.CookieClearingLogoutHandler;
 import org.springframework.security.web.csrf.CookieCsrfTokenRepository;
+import org.springframework.security.web.csrf.CsrfFilter;
 import org.springframework.security.web.header.writers.ContentSecurityPolicyHeaderWriter;
 import org.springframework.security.web.header.writers.DelegatingRequestMatcherHeaderWriter;
 import org.springframework.security.web.savedrequest.HttpSessionRequestCache;
 import org.springframework.security.web.util.matcher.AntPathRequestMatcher;
+import org.springframework.security.web.util.matcher.OrRequestMatcher;
 
 
 /**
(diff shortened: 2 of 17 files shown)


pass-core-main/src/test/java/org/eclipse/pass/main/security/AccessControlTest.java
```diff
@@ -229,14 +234,65 @@ public void testCreateUpdateDeleteSubmissionAsShibUser() throws IOException, JSO
...
 
             check(response, 204);
         }
     }
 
+    @Test
+    public void testCreateUpdateDeleteSubmissionAsShibUserWithBadCsrfToken() throws IOException, JSONException {
+        doSamlLogin();
+
+        {
```
rpoet-jh, line 251: To be consistent with other tests, I would recommend removing the Block syntax and just name variables appropriately.

pass-core-main/src/test/java/org/eclipse/pass/main/security/AccessControlTest.java
```diff
@@ -542,14 +612,86 @@ public void testCreateUpdateDeleteGrantAsBackend() throws IOException, JSONExcep
...
+        set_attribute(grant, "projectName", "backend test");
+
+        // Save our own CSRF token
+        client.cookieJar().saveFromResponse(base_url,
+                List.of(Cookie.parse(base_url, "XSRF-TOKEN=moo")));
+
+        assertEquals("moo", getCsrfToken());
+
+        // Create a grant with a made up CSRF token
+        {
```
rpoet-jh, line 680: Block not needed here.

pass-core-main/src/test/java/org/eclipse/pass/main/security/AccessControlTest.java
```diff
@@ -542,14 +612,86 @@ public void testCreateUpdateDeleteGrantAsBackend() throws IOException, JSONExcep
...
+        String url = getBaseUrl() + "data/grant";
+
+        // Save our own CSRF token
+        assertNull(get_cookie("XSRF-TOKEN"));
+        client.cookieJar().saveFromResponse(base_url,
+                List.of(Cookie.parse(base_url, "XSRF-TOKEN=moo")));
+        assertNotNull(get_cookie("XSRF-TOKEN"));
+
+        // Create a grant with CSRF token
+        {
```
rpoet-jh, line 640: To be consistent with other tests, I would recommend removing the Block syntax and just name variables appropriately.

rpoet-jh (commented):
Great job, Mark!  The spring security configuration looks like it pretty much matches what is in the docs for SPA CSRF which is good.  I only had a few minor comments.

The question re: what to do about what to do about requests coming from the `backend user (stateless basic auth)` and CSRF, I will do some more reading on this, but we can also talk about this if you want to try and come to a reasonable decision.

pass-core-main/src/test/java/org/eclipse/pass/main/security/AccessControlTest.java
markpatton: Pretty much all the tests are using block syntax.

pass-core-main/src/main/java/org/eclipse/pass/main/security/CsrfCookieFilter.java
```diff
@@ -0,0 +1,25 @@
...
+
+import jakarta.servlet.FilterChain;
+import jakarta.servlet.ServletException;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
+import org.springframework.security.web.csrf.CsrfToken;
+import org.springframework.web.filter.OncePerRequestFilter;
+
+/**
+ * Ensure CSRF token cookie is added.
```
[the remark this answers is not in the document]
markpatton: Done.

pass-core-main/src/main/java/org/eclipse/pass/main/security/SecurityConfiguration.java
```diff
@@ -77,10 +77,16 @@ public class SecurityConfiguration {
     @Bean
     public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
         // Disable unused functionality
-        http.csrf(CsrfConfigurer::disable);
         http.formLogin(FormLoginConfigurer::disable);
         http.anonymous(AnonymousConfigurer::disable);
 
+        // Enable CSRF protection using a cookie to send the token
+        // Make sure the cookie value can be parsed when returned in a header
+        // Do not protect /logout so it can be triggered with GET
```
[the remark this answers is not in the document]
markpatton: Done

pass-core-main/src/main/java/org/eclipse/pass/main/security/SpaCsrfTokenRequestHandler.java
```diff
@@ -0,0 +1,49 @@
...
+
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
+import org.springframework.security.web.csrf.CsrfToken;
+import org.springframework.security.web.csrf.CsrfTokenRequestAttributeHandler;
+import org.springframework.security.web.csrf.CsrfTokenRequestHandler;
+import org.springframework.security.web.csrf.XorCsrfTokenRequestAttributeHandler;
+
+/**
+ * Handle case of single page application which puts CSRF token cookie value into header.
```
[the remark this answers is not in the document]
markpatton: Done.

pass-core-main/src/main/java/org/eclipse/pass/main/security/SpaCsrfTokenRequestHandler.java
```diff
@@ -0,0 +1,49 @@
...
+    public String resolveCsrfTokenValue(HttpServletRequest request, CsrfToken csrfToken) {
+        /*
+         * If the request contains a request header, use CsrfTokenRequestAttributeHandler
+         * to resolve the CsrfToken. This applies when a single-page application includes
+         * the header value automatically, which was obtained via a cookie containing the
+         * raw CsrfToken.
+         */
+        String token = request.getHeader(csrfToken.getHeaderName());
+
+        if (token != null && !token.isEmpty()) {
```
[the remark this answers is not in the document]
markpatton: Done. That will prevent a csrf token from consisting entirely of whitespace, which certainly won't matter.

pass-core-main/src/test/java/org/eclipse/pass/main/security/AccessControlTest.java
markpatton: Needed because of how the http requests were structured in blocks. It could be rewritten without blocks, but to me this seems more readable.

pass-core-main/src/test/java/org/eclipse/pass/main/security/AccessControlTest.java
markpatton: Needed because of how the http requests were structured in blocks. It could be rewritten without blocks, but to me this seems more readable. And a bunch of tests are doing the same thing.

rpoet-jh:
I'm ready to approve, but I know there are still some pending issues on the UI/tests side.  I will approve once those issue have been fixed jic there are changes need in pass-core.

rpoet-jh (approved):

PR #91, opened 2024-08-26
Change repackage configuration

rpoet-jh:
This PR makes the following changes to the `repackage` configuration:

- Remove the `finalName` configuration.  It is read-only and should not be changed any longer: https://github.com/spring-projects/spring-boot/issues/16202#issuecomment-500208885
- Set `attach` as false so the repackaged jar file is not deployed.  If this works like I think it will, it may speed up the deploy goal since the repackaged jar file will not be deployed to sonatype snapshot/release.
- Removed `pluginManagement` from pass-core root pom since pass-core-main is the only module using repackage.

### DIFF
--- a/pass-core-main/Dockerfile
+++ b/pass-core-main/Dockerfile
@@ -2,7 +2,7 @@ FROM eclipse-temurin:17.0.8_7-jre-jammy
 
 WORKDIR /app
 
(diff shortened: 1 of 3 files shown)


tsande16 (approved):
`mvn clean install` runs successfully. Looks good!

PR #92, opened 2024-08-26
Fix handling of SAML optional attributes

markpatton:
Some SAML attributes, employee id and affiliation, are optional. A bug was requiring them instead.

### DIFF
--- a/pass-core-main/src/test/java/org/eclipse/pass/main/security/PassAuthenticationFilterTest.java
+++ b/pass-core-main/src/test/java/org/eclipse/pass/main/security/PassAuthenticationFilterTest.java
@@ -5,16 +5,22 @@ import static org.junit.jupiter.api.Assertions.assertNotEquals;
 
 import java.io.IOException;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


pass-core-main/src/test/java/org/eclipse/pass/main/security/PassAuthenticationFilterTest.java
```diff
@@ -65,4 +71,61 @@ public void testLoginUpdatesUser() throws IOException {
...
+        expected.setUsername("thomassubmitter987654321@johnshopkins.edu");
+        expected.setRoles(List.of(UserRole.SUBMITTER));
+
+        User user = passAuthFilter.parseUser(attributes);
+
+        assertEquals(expected, user);
+    }
+
+    @Test
+    public void testParseUserMissingEmployeeIdAndAffiliation() {
```
rpoet-jh, line 106: Is there are test for missing required attributes?

markpatton: No. That is a good point.

markpatton: Added a test for that case.

rpoet-jh (approved):

PR #93, opened 2024-10-07
Add option to load app props from AWS Parameter Store

rpoet-jh:
This PR adds the option to load application properties defined in application.yaml from and AWS Parameter Store.

The PR contains the following changes:

- Rename `PASS_CORE_BACKEND_USER` and `PASS_CORE_BACKEND_PASSWORD` to `PASS_CORE_USER` and `PASS_CORE_PASSWORD` respectively. This will set up usage of these by other services/data loaders in PASS.
- Add spring AWS Parameter Store dep to implement app config environment integration with param store.

The way it works is by using the `spring.config.import` to specify to load app props from the AWS parameter store.  See the [AwsParamStoreConfigTest.properties method](https://github.com/eclipse-pass/pass-core/pull/93/files#diff-2778635f5442ce8d706b8764e54a768f91758c92978b39ff040029791bba55d0R55) for an example.

https://docs.awspring.io/spring-cloud-aws/docs/3.0.0/reference/html/index.html#spring-cloud-aws-parameter-store

Note that this PR needs to be approved and merged before the other PR in pass-support.  This is because of the pass core user/password env var name change.


rpoet-jh:
The acceptance tests are failing because pass-docker needs to be updated to use the new pass core user env names. I think this is a case where we will have to merged one of the PRs first to get the other to pass. I'd say to merge this one first, then we can re-run the pass-docker PR to ensure it works.

markpatton (approved):
Looks good. As mentioned pass-documentation needs to be updated wrt the env variable change.

PR #94, opened 2024-10-14
Modify DOI service to not return entries without a URL

markpatton:
This PR updates the DOI service such that entries without a URL are not returned. In addition the returned filenames are parsed from the URL in a more reliable way so that we do not return filenames like `blah.pdf?render=pdf`.


rpoet-jh (approved):
Looks good.

PR #95, opened 2024-10-15
Remove stacktrace from error responses

rpoet-jh:
This will fix the issue with showing the stacktrace on the Whitelabel error page for HTML clients.  

Note that the stacktrace was also being included in error responses for JSON clients.  With this PR, the stacktrace will no longer be in the error response for JSON clients as well.

The Whitelabel Error Page now looks like:

Screenshot 2024-10-15 at 4 07 31 PM


markpatton (approved):
Makes sense. I think I had added that for debugging, but it can always be turned back on when needed.

PR #96, opened 2024-10-24
Clean default application configuration and property values

rpoet-jh:
The following changes are in this PR:

- Change application.yaml so the default configuration is setup for production use.  This means no default values for sensitive or institutional specific configuration properties.  Note there is no production profile now, it is the default configuration.
- Create application-test.yaml that has configuration property values for local running and testing.  This includes moving the `src/main/resources/saml2` directory to `src/test/resources/saml2`.
- Remove docker compose files since they seemed outdated.  pass-docker can be used to run pass-core.
- Update README to reflect configuration changes.


markpatton (approved):
This is a really good cleanup. I tested locally and it behaved as expected when choosing the default (produciton) or test profiles.

PR #99, opened 2024-10-30
Move pass-core-main test configuration

rpoet-jh:
To new module named `pass-core-test-config` that be used by other projects for ITs.  A new pass-core-test-config-<release>.jar file is created/published that can be used by pass-support ITs that start pass-core for ITs. 

https://github.com/eclipse-pass/main/issues/1071


markpatton (approved):
I think having a separate test config module is a good way to go. Tests pass for me locally.

PR #100, opened 2024-11-04
Upgrade dependencies to latest versions

rpoet-jh:
This PR upgrades pass-core dependencies to latest versions.  There were only a couple breaking changes:

- Elide EntityDictionary add an optional param in the constructor.  Set to null assuming we don't want to set an `IdObfuscator`.
- OCFL changed types of S3 objects used to TransferManager and Async clients.

Also, the Docker FROM image was updated to latest eclipse-temurin JRE 17 jammy image.


rpoet-jh:
All tests pass locally, but they are failing for some reason in the gh workflow action.  Investigating.

markpatton (approved):
Looks good. The tests pass for me locally.

PR #104, opened 2024-11-25
Fix publication date handling in metadata

markpatton:
* Make publication date required for JScholarship.
* Treat publication date as a date not a string so a picker will popup
* Use localized date format for date picker display.

The date will be stored in the metadata as a local ISO 8601 date format (no timezone). I spent a while trying to get the timezone returned, but to no avail. It should be possible, but may have to wait for an updated alpaca and date picker.

Required prs:
  * https://github.com/eclipse-pass/pass-acceptance-testing/pull/28
  * https://github.com/eclipse-pass/pass-ui/pull/1297


rpoet-jh:
@markpatton this PR (and a few of the other ones) is still in Draft, is this ready for review?

pass-core-metadataschema-service/src/main/resources/schemas/jhu/global.json
```diff
@@ -264,8 +264,16 @@
             },
             "publicationDate": {
                 "type": "date",
-                "label": "Publication Date  <small class=\"text-muted\">(optional)</small>",
+                "label": "Publication Date  <small class=\"text-muted\"></small>",
                 "hidden": false,
+                "helpersPosition": "above",
```
rpoet-jh, line 269: Looks like there are matching files in the `tests/resources`, should they be updated too?

markpatton: That is a good question. I'm not sure that particular change would be needed for tests. Another question is, if the test resources are supposed to duplicate the main resources, why have the test resources at all? I will experiment a bit.

PR #107, opened 2024-12-21
Add JaCoCo for Code Coverage Reports in Pass Core

tsande16:
This PR will add JaCoCo (Java Code Coverage) to the pass-core project. It is mostly the default configuration for JaCoCo. The reports are aggregated into the `jacoco-aggregate-report` module. In my research this is the approach to take when setting up an aggregate report to be consumed by a report service like SonarQube.  

- Report can be found in `target/site/jacoco/index.html` in the `jacoco-aggregate-report` module
   - Reports on individual classes are found in subfolders such as `target/site/jacoco/org.eclipse.pass.support.client` etc. These can be navigated to by using the `index.html`
- I removed the coverage ratio so it wouldn't fail the build as we determined in our last status meeting, the first iteration will be just to view the reports and then determine what type of metrics to fail a build.
- There was some minor javadoc added to the FileService.

To generate the reports:
- run `mvn verify`

### DIFF
--- a/jacoco-aggregate-report/pom.xml
+++ b/jacoco-aggregate-report/pom.xml
@@ -12,9 +12,6 @@
     <packaging>pom</packaging>
 
     <properties>
-        <maven.compiler.source>17</maven.compiler.source>
-        <maven.compiler.target>17</maven.compiler.target>
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     </properties>
 
     <dependencies>
--- a/jacoco-aggregate-report/pom.xml
+++ b/jacoco-aggregate-report/pom.xml
@@ -38,6 +38,21 @@
             <artifactId>pass-core-object-service</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.pass</groupId>
+            <artifactId>pass-core-policy-service</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.pass</groupId>
+            <artifactId>pass-core-user-service</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.pass</groupId>
+            <artifactId>pass-core-usertoken</artifactId>
+            <version>${project.version}</version>
+        </dependency>
     </dependencies>
 
     <build>
--- a/jacoco-aggregate-report/pom.xml
+++ b/jacoco-aggregate-report/pom.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
+
     <parent>
         <groupId>org.eclipse.pass</groupId>
         <artifactId>pass-core</artifactId>
@@ -8,6 +9,7 @@
     </parent>
 
     <artifactId>jacoco-aggregate-report</artifactId>
+    <packaging>pom</packaging>
 
     <properties>
         <maven.compiler.source>23</maven.compiler.source>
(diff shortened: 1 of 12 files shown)


tsande16:
@markpatton @rpoet-jh Where should we put the documentation for this? Perhaps it should have it's own section in the `Developer Documentation` section?

jacoco-aggregate-report/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.eclipse.pass</groupId>
+        <artifactId>pass-core</artifactId>
+        <version>1.14.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>jacoco-aggregate-report</artifactId>
```
rpoet-jh, line 11: Since `pass-core-main` depends on all other modules already, did you try putting this jacoco pom config (`report-aggregate`) into the `pass-core-main` module pom?  I wonder if that may be better if it works so that we don't add another module to `pass-core`?  I know there has been talk about wanting to consolidate the modules into one for pass-core.

jacoco-aggregate-report/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+    <packaging>pom</packaging>
+
+    <properties>
+        <maven.compiler.source>23</maven.compiler.source>
```
rpoet-jh, line 15: `pass-core` uses java 17 atm.

rpoet-jh (commented):
Good job, Tim.  Just a couple comments.

jacoco-aggregate-report/pom.xml
markpatton: Yes. I'd like to eventually consolidate all the pass-core modules. I don't think the separation has much point. I agree about using pass-core main if possible. In fact I was wondering if it would work in the pass root module to just cover everything.

jacoco-aggregate-report/pom.xml
tsande16: Good catch. It's interesting that IntelliJ generated that even though my project is set to 17.

jacoco-aggregate-report/pom.xml
tsande16: @rpoet-jh I will give this a shot and see what happens. I followed this setup from [JaCoCo](https://github.com/jacoco/jacoco/tree/v0.7.8/jacoco-maven-plugin.test/it/it-report-aggregate)

From my understanding the report aggregation needs to be it's own separate module. When I originally tried doing this I tried setting the project POM to be the aggregator, but that didn't work. From my reading of the docs, if it was setup in `pass-core-main` it wouldn't generate a report for pass-core-main, but would only collect from the other submodules and place the aggregated report in `pass-core-main`.

jacoco-aggregate-report/pom.xml
tsande16: Actually the only other POM that has it is in the `pass-core-policy` submodule.

jacoco-aggregate-report/pom.xml
tsande16: @markpatton @rpoet-jh This is what happens when I move it the aggregation to `pass-core-main`. It won't create a report for `pass-core-main`. I've tried playing around with different phases, scopes, with/without the agent in `pass-core-main`. 

image

markpatton: @rpoet-jh @tsande16 No. Let's go for it.

tsande16:
@rpoet-jh Following up from our zoom convo from earlier. I tested the question you brought up to see if report aggregation works with tests that exist in another module.

Setup: The FileService tests exist in `pass-core-main` module. I picked this method which was missing coverage for this one branch:

image

I added a test for the getFileById() exception logic with a missing ID or file name:

    @Test
    void testGetFileByIdMissingFileId() {
        ResponseEntity<?> response = passFileServiceController.getFileById("", "file.txt");

        assertEquals(400, response.getStatusCodeValue());
        assertEquals("File ID not provided to get a file.", response.getBody());
    }
    
And it now shows it as covered in the JaCoCo Report:

image

rpoet-jh (approved):

PR #108, opened 2025-01-06
Remove graphiql resources

rpoet-jh:
Resolves SQ security hotspot.  Graphql is disabled in pass-core, so these files are not needed.


markpatton (approved):
Looks good.

PR #109, opened 2025-01-09
Set USER for docker

rpoet-jh:
This PR updates the Dockerfile so the pass-core docker container runs with a non-privileged user.

I have tested this locally in pass-docker and acceptance tests pass.

### DIFF
--- a/pass-core-main/Dockerfile
+++ b/pass-core-main/Dockerfile
@@ -6,8 +6,14 @@ COPY target/pass-core-main-*-exec.jar pass-core-main.jar
 COPY entrypoint.sh .
 
 RUN apt update \
-    && apt install -y curl \
+    && apt --no-install-recommends install -y curl \
     && chmod +x entrypoint.sh \
     && rm -rf /var/lib/apt/lists/*
+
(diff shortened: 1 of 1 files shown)


pass-core-main/Dockerfile
```diff
@@ -6,8 +6,14 @@ COPY target/pass-core-main-*-exec.jar pass-core-main.jar
 COPY entrypoint.sh .
 
 RUN apt update \
-    && apt install -y curl \
+    && apt --no-install-recommends install -y curl \
     && chmod +x entrypoint.sh \
     && rm -rf /var/lib/apt/lists/*
 
+RUN groupadd -g 1432 passcoregroup && \
```
markpatton, line 13: Is there a need to set a specific ids for the group and user?

rpoet-jh: Seems it is best practice when executing `groupadd/useradd`, see `Use custom user and group IDs` in https://www.docker.com/blog/understanding-the-docker-user-instruction/

markpatton (approved):
Looks good. Runs locally for me.

PR #110, opened 2025-01-13
Enable CSRF for /logout

rpoet-jh:
This PR enables CSRF requirement for the `/logout` endpoint.

Note the removal of `http.saml2Logout` from the security configuration.  This is needed so the SAML SLO logout doesn't execute when a POST is sent to `/logout`.  I assume this is what we want because in the previous configuration, SAML SLO logout was essentially disabled because it only executes for POST requests, so it was being skipped for the GET `/logout` requests.  Let me know if I missed something with this assumption.

There will also be a `pass-ui` PR that must be merged with this PR. https://github.com/eclipse-pass/pass-ui/pull/1294


rpoet-jh:
To test in pass-docker, build pass-core and pass-ui (on branch for https://github.com/eclipse-pass/pass-ui/pull/1294).  Update pass-docker to run pass-ui:latest as usual.

PR #112, opened 2025-01-21
Clean up of dependencies with convergence issues

rpoet-jh:
This PR declares all the dependencies previously excluded from dependency convergence, so that the version used by the application is predictable.  The highest version of each dependency was declared.


markpatton (approved):
Looks good. This tightens up the dependencies.

PR #118, opened 2025-02-19
Upgrade dependencies

rpoet-jh:
Upgrade dependencies to resolve Security Vulnerabilities identified in https://github.com/eclipse-pass/main/issues/1109.

The only Security Vulnerability left is the h2 dependency vulnerability.  Looks like this is a false positive and being flagged by bad info in OSS: https://github.com/OSSIndex/vulns/issues/277.  Not sure why it is showing up still, but pass-core h2 version is `2.3.232`, which based on a couple sites I checked looks clean: https://deps.dev/maven/com.h2database%3Ah2/2.3.232

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -76,13 +76,13 @@
 
   <properties>
     <maven-dependency-plugin.version>3.8.1</maven-dependency-plugin.version>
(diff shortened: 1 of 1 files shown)


markpatton (approved):
Makes sense.

PR #120, opened 2025-02-20
Update doi service to transform JATS XML abstracts to HTML

markpatton:
The CrossRef object returned by the DOI service is modified such that JATS XML is transformed to HTML using XSLT stylesheets.

In order to test, run locally in pass-docker and do a submission. Enter a DOI and on the metadata page, check that the abstract is HTML.


rpoet-jh (approved):
Good job, Mark!  Tests pass for me locally.  I also ran some tests in pass-docker with DOIs that had JATS formatted abstracts.  They showed as HTML in the Details-Abstract form field.

PR #123, opened 2025-04-21
Enable spring cloud AWS s3

rpoet-jh:
So that pass-core can use spring cloud aws to use S3 if enabled.


markpatton (approved):
Looks good.

PR #124, opened 2025-04-29
Remove no longer used metadata schema service

markpatton:
Acceptance tests will fail until https://github.com/eclipse-pass/pass-ui/pull/1301 is merged.


markpatton:
One of the proxy acceptance tests is failing consistently in the action, but it passes for me when run locally.

rpoet-jh (approved):
Looks good.

PR #126, opened 2025-06-04
Change DOI lookup Journal update rules

### LINKED ISSUE #1189: Update PASS Journal.journalName on DOI lookup
# What? 

When a DOI is looked up in pass-core, if the associated Journal exists in PASS, update the journalName if it is different.

# Why?

To keep journal name up to date in PASS.

# How?

Check if journal name in PASS is different from DOI info and update.

# Acceptance Criteria 

Journal.journalName stays up to date with DOI journal name.

# Related Issues

rpoet-jh:
So that on DOI lookup, the associated Journal will only be updated if it is not updated by the journal loader. Also added updating of `Journal.journalName` if updates occur.


markpatton (approved):
Looks good. The test updates are nice too.

PR #128, opened 2025-08-19
Bump bouncycastle version

rpoet-jh:
To make dependabot update so CI passes: https://github.com/eclipse-pass/pass-core/pull/127


markpatton (approved):
Looks good.

PR #130, opened 2025-09-24
Add pom elements in child modules for publish

rpoet-jh:
Apparently it is now required by sonatype central repo to duplicate a set of POM attributes in child POM files in order to publish.


markpatton (approved):
Looks good. A bit annoying to have to duplicate the information. Let's see if it makes sonatype happy.

PR #132, opened 2025-11-17
Clean POMs and upgrade bouncycastle

rpoet-jh:
This PR removes the POM attributes from the child POMs that are inherited from the parent POM.

It also ugprades bouncycastle to resolve this dependabot PR https://github.com/eclipse-pass/pass-core/pull/131


markpatton (approved):
Looks good.

PR #134, opened 2026-03-12
Normalize error handling for DOI service

markpatton:
- The disabled tests for the doi service were not running because error handling was a little fragile. This pr normalizes the error handling.
- The code which prevented simultaneous requests for the same DOI has also been removed because it is not needed.
- Some character encoding issues with responses fixed.
- I also had to rejigger some tests and add more unit testing to make Sonarqube happy

### DIFF
--- a/pass-core-doi-service/src/main/java/org/eclipse/pass/doi/service/PassDoiServiceController.java
+++ b/pass-core-doi-service/src/main/java/org/eclipse/pass/doi/service/PassDoiServiceController.java
@@ -19,14 +19,13 @@ package org.eclipse.pass.doi.service;
 import java.io.IOException;
 import java.io.Writer;
 
-import com.yahoo.elide.RefreshableElide;
 import jakarta.json.Json;
 import jakarta.json.JsonObject;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RestController;
 
@@ -44,29 +43,10 @@ public class PassDoiServiceController {
     private final ExternalDoiService xrefDoiService;
     private final ExternalDoiService unpaywallDoiService;
 
-    /**
-     * @param refreshableElide the RefreshableElide
-     */
-    @Autowired
-    public PassDoiServiceController(RefreshableElide refreshableElide) {
-        this.elideConnector = new ElideConnector(refreshableElide);
-        this.externalDoiServiceConnector = new ExternalDoiServiceConnector();
-        this.xrefDoiService = new XrefDoiService();
-        this.unpaywallDoiService = new UnpaywallDoiService();
-    }
-
-    /**
-     * Constructor for testing, allows injection of mock connectors and services.
-     *
-     * @param elideConnector
-     * @param externalDoiServiceConnector
-     * @param xrefDoiService
-     * @param unpaywallDoiService
-     */
-    PassDoiServiceController(ElideConnector elideConnector,
-                             ExternalDoiServiceConnector externalDoiServiceConnector,
-                             ExternalDoiService xrefDoiService,
-                             ExternalDoiService unpaywallDoiService) {
+    public PassDoiServiceController(ElideConnector elideConnector,
+                            ExternalDoiServiceConnector externalDoiServiceConnector,
+                            @Qualifier("xrefDoiService") ExternalDoiService xrefDoiService,
+                            @Qualifier("unpaywallDoiService") ExternalDoiService unpaywallDoiService) {
         this.elideConnector = elideConnector;
         this.externalDoiServiceConnector = externalDoiServiceConnector;
         this.xrefDoiService = xrefDoiService;
--- a/pass-core-doi-service/src/main/java/org/eclipse/pass/doi/service/PassDoiServiceController.java
+++ b/pass-core-doi-service/src/main/java/org/eclipse/pass/doi/service/PassDoiServiceController.java
@@ -17,7 +17,7 @@
 package org.eclipse.pass.doi.service;
 
 import java.io.IOException;
-import java.io.OutputStream;
+import java.io.Writer;
 
 import com.yahoo.elide.RefreshableElide;
 import jakarta.json.Json;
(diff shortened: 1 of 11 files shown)


pass-core-doi-service/src/main/java/org/eclipse/pass/doi/service/PassDoiServiceController.java
```diff
@@ -46,13 +47,32 @@ public class PassDoiServiceController {
     /**
      * @param refreshableElide the RefreshableElide
      */
+    @Autowired
     public PassDoiServiceController(RefreshableElide refreshableElide) {
         this.elideConnector = new ElideConnector(refreshableElide);
```
rpoet-jh: I would recommend using Spring to create and wire up these DOI classes.  

You could add the `@Service` annotation to `XrefDoiService` and `UnpaywallDoiService`, and `@Component` to `ExternalDoiServiceConnector` and `ElideConnector`. Then just pass them all into this constructor and set.

By doing this, you could then use `@MockitoBean` in tests to mock as required, and the constructor below would not be needed.

pass-core-doi-service/src/test/java/org/eclipse/pass/doi/service/PassDoiServiceControllerTest.java
```diff
@@ -0,0 +1,262 @@
...
+
+import jakarta.json.Json;
+import jakarta.json.JsonObject;
+import jakarta.json.JsonReader;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class PassDoiServiceControllerTest {
```
rpoet-jh: Looks like some of these tests are actually covered in the IT in pass-core-main. For example, `testGetXrefMetadata_ServiceError_404` looks to be covered with this test  https://github.com/eclipse-pass/pass-core/blob/65ea36e4b90f71ec03a3acbaf0fd19cf239b440e/pass-core-main/src/test/java/org/eclipse/pass/doi/service/DoiServiceTest.java#L140

Normally, Controller tests are ITs. I suggest looking over these tests and seeing if they are already covered in DoiServiceTest. There may be some tests that cannot be tested easily as an integration test, those can stay in this class.

pass-core-doi-service/src/test/java/org/eclipse/pass/doi/service/ExternalDoiServiceConnectorTest.java
```diff
@@ -1,94 +1,110 @@
...
     @Test
-    public void testUnpaywallLookup() {
-        String realDoi = "10.4137/cmc.s38446"; //"10.1038/nature12373";
-        JsonObject blob = underTest.retrieveMetadata(realDoi, unpaywallService);
+    void testRetrieveMetadataNotFound() throws IOException {
+        try (MockWebServer server = new MockWebServer()) {
+            server.enqueue(new MockResponse().setResponseCode(404));
 
-        JsonObject object = JsonTestObjectsUtil.unpaywallTestJsonObject();
+            ExternalDoiServiceConnector underTest = new ExternalDoiServiceConnector(okhttpClient);
```
rpoet-jh: Instead of having the constructor in `ExternalDoiServiceConnector` to set the okHttpClient that is used only for tests, you can use `ReflectionTestUtils.setField(underTest, "client", okhttpClient);` to set it after construction.

pass-core-doi-service/src/main/java/org/eclipse/pass/doi/service/UnpaywallDoiService.java
```diff
@@ -20,18 +20,21 @@
...
 import jakarta.json.JsonObject;
 import jakarta.json.JsonValue;
+import org.springframework.stereotype.Service;
 
 /**
  * The UnpaywallDoiService class is an implementation of the ExternalDoiService abstract class to interface with
  * the Unpaywall API. The Unpaywall API is a RESTful API that returns JSON metadata for a given DOI. More information
  * about the Unpaywall API can be found here: <a href="https://unpaywall.org/products/api">Unpaywall API</a>
  */
+@Service()
```
rpoet-jh: Parens not needed

pass-core-doi-service/src/main/java/org/eclipse/pass/doi/service/XrefDoiService.java
```diff
@@ -44,6 +45,7 @@
  * address. The default email address used by is pass@jhu.edu and can be overridden by setting the environment variable
  * PASS_DOI_SERVICE_MAILTO
  */
+@Service()
```
rpoet-jh: Parens not needed.

rpoet-jh (commented):
Good job on the Spring IOC changes. I reviewed all the code after that, just a few more comments.

pass-core-doi-service/src/test/java/org/eclipse/pass/doi/service/ExternalDoiServiceConnectorTest.java
markpatton: Rejiggered this to avoid the need for the constructor.

pass-core-doi-service/src/test/java/org/eclipse/pass/doi/service/PassDoiServiceControllerTest.java
markpatton: The problem is the permutations with the external services. The integration test can only really test the success path since it hits the external services. If you start mocking the external services, then its not really an integration test anymore. So I just created a separate unit test.

The unit test gives us a way to test the failure paths of the external service integrations like Sonarqube is demanding. There is some test redundancy between the two approaches, but it doesn't seem like a big deal. I will add some documentation to the tests so it is less confusing.

markpatton:
@rpoet-jh 

Made some updates. See what you think.

rpoet-jh:
> @rpoet-jh
> 
> Made some updates. See what you think.

Looks good. Only comment is I think you can remove the constructor `ExternalDoiServiceConnector(OkHttpClient client)` now?

rpoet-jh:
@markpatton oh yeah, to clean up a bunch of the sq warnings, you can remove the `public` scope from class and methods from the `PassDoiServiceControllerTest` test.

rpoet-jh:
> Only comment is I think you can remove the constructor `ExternalDoiServiceConnector(OkHttpClient client)` now?

@markpatton you may have missed this small comment.

markpatton:
@rpoet-jh 

I thought I had all of them! I see they didn't trigger the failure mode. Cleaned up. I wish there was a way to get a list of all the issues on the command line.

rpoet-jh:
> @rpoet-jh
> 
> I thought I had all of them! I see they didn't trigger the failure mode. Cleaned up. I wish there was a way to get a list of all the issues on the command line.

@markpatton did you push the commit? I don't see it.

rpoet-jh (approved):
One of the acceptance tests failed, I think there is some flakiness in that test, I saw it fail a couple days back too. Not caused by this PR, approved.

PR #135, opened 2026-04-06
Update dependencies for secops

rpoet-jh:
This PR upgrades a couple dependencies to resolve Critical CVEs in those third party dependencies.

The spring security one is more critical. Activemq artemis is only used for testing purposes.


markpatton (approved):
Looks good. Logging dependencies are always a mess!

